PR #213, opened 2025-09-30
engineering: Do not clear TPM 2.0 on clean install by default, but introduce `encryption.clearTpmOnInstall` option

ndubchak:
# 🔍 Description
 
This PR changes the current logic for creating encrypted devices, by **not** clearing the TPM 2.0 device by default on clean install, unless `encryption.clearTpmOnInstall` is set to true.

Green E2E runs:
1. BM: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=941820&view=results
2. trident-pr-e2e: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=941328&view=results

# 🤔 Rationale

The ZTP team has a multi-boot scenario, where the TPM 2.0 device is already in use on Azure Local: the TPM is used for disk encryption on the ROE data partitions, and clearing the TPM breaks the transition between the ROE and the Target OS. Considering this, we decided to **not** clear the TPM 2.0 on clean install by default. However, this behavior can still be enabled, by setting `cleanTpmOnInstall: true` in the `storage.encryption` config on clean install. This might be needed for some development or testing scenarios, where the TPM 2.0 needs to be cleared, e.g. to avoid the DA lockout mode on repeated provisioning attempts.

Related ADO items:

1. https://dev.azure.com/mariner-org/polar/_workitems/edit/15420/
3. https://dev.azure.com/mariner-org/polar/_workitems/edit/15443/

### DIFF
--- a/crates/trident_api/src/config/host/storage/encryption.rs
+++ b/crates/trident_api/src/config/host/storage/encryption.rs
@@ -101,6 +101,18 @@ pub struct Encryption {
     /// More encryption flows, with additional PCR options, will be added in the future.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub pcrs: Vec<Pcr>,
+
+    /// Optional parameter that determines whether the TPM 2.0 device will be cleared on clean
+    /// install. By default, it is set to false. If set to true, Trident will clear the TPM 2.0
+    /// device on install. TPM cannot be cleared on A/B updates.
+    ///
+    /// Clearing the TPM 2.0 device will remove all keys and data from the TPM 2.0 device. This
+    /// operation is irreversible and will result in data loss. However, this option might be
+    /// needed to ensure that the TPM 2.0 is in a known state; to avoid entering the DA (Direct
+    /// Attack) lockout mode on repetitive provisioning attempts, e.g. during testing and
+    /// development.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub clear_tpm_on_install: Option<bool>,
 }
 
 /// A LUKS2-encrypted volume configuration.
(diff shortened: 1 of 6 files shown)


tests/e2e_tests/target-configurations.yaml
```diff
@@ -7,22 +7,22 @@ bareMetal:
...
-      - root-verity
-      - simple
-      - split
-      - usr-verity
-      - usr-verity-raid
+      #- root-verity
+      #- simple
+      #- split
+      #- usr-verity
+      #- usr-verity-raid
```
ndubchak, line 25: TODO: Need to un-do these changes, but should probably first run the BM `rerun` test, since it failed here when we made changes to go public: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=939090&view=results

crates/trident_api/src/config/host/storage/encryption.rs
```diff
@@ -101,6 +101,18 @@ pub struct Encryption {
     /// More encryption flows, with additional PCR options, will be added in the future.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub pcrs: Vec<Pcr>,
+
+    /// Optional parameter that determines whether the TPM 2.0 device will be cleared on clean
+    /// install. By default, it is set to false. If set to true, Trident will clear the TPM 2.0
+    /// device on install. TPM cannot be cleared on A/B updates.
+    ///
+    /// Clearing the TPM 2.0 device will remove all keys and data from the TPM 2.0 device. This
+    /// operation is irreversible and will result in data loss. However, this option might be
```
fintelia, line 110: nit: will -> could

PR #221, opened 2025-10-01
engineering: Include PCR 7 into pcrlock policy for non-containerized UKI target OS

ndubchak:
# 🔍 Description
 
This PR un-comments PCR 7 in our encryption E2E HCs, so that PCR 7 can be included into the pcrlock policy for UKI images. Also, the HC validation logic is modified so that now all three supported PCRs: 4, 7, 11, are allowed for UKI images.

Moreover, this PR does the following:

1. [DONE] Validate that if PCR 7 is included for a target UKI OS, then SecureBoot is enabled and we're NOT running inside a container
2. [DONE] Remove PCR 7 from our E2E HCs when container tests are run
3. [DONE] Remove default PCR (PCR 7) and require PCRs to be specified in every HC
4. [DONE] Exclude PCR 7 on BM E2E tests since `SecureBoot` is not enabled on the BM machines yet

# 🤔 Rationale

This PR is needed for the PCR encryption work.

TODO:

- https://dev.azure.com/mariner-org/polar/_workitems/edit/15566
- https://dev.azure.com/mariner-org/polar/_workitems/edit/14455/
- https://dev.azure.com/mariner-org/polar/_workitems/edit/13346/


crates/trident_api/src/config/host/storage/encryption.rs
```diff
@@ -81,25 +81,46 @@ pub struct Encryption {
...
-    /// - 11, or `kernel-boot`,
-    /// - 4 and 11.
+    /// - 7, or `secure-boot-policy`
+    ///
+    /// ### UKI Target OS
+    ///
+    /// When doing a clean install of a UKI target OS image, the following options are valid:
+    ///
+    /// - 4, or `boot-loader-code`
+    ///
```
bfjelds, line 99: curious why the line between each list item?

ndubchak: It doesn't render as a bullet point list otherwise, I think. At least when I click on "View", it doesn't; does Docusaurus do it?

bfjelds (approved):

crates/trident_api/src/config/host/storage/encryption.rs
bfjelds: looking with docusaurus locally now

PR #270, opened 2025-10-16
engineering: Installer ISO definition

alejandro-microsoft:
# 🔍 Description

- ISO definition to produce the ISO for attended installation. 
- Make rule to generate the ISO.
- Disabled the EULA view and requirement (this can be re-enabled when needed).


tests/images/azl-installer/iso/files/EULA.txt
```diff
@@ -0,0 +1,55 @@
+MICROSOFT SOFTWARE LICENSE TERMS
```
bfjelds: where did this come from?  do we need any approvals?

tests/images/azl-installer/mos/mos.yaml
```diff
@@ -0,0 +1,81 @@
+storage:
```
bfjelds: is this PR just about the changes under tests/images/azl-installer?  seems like the other changes in this PR (tools/installer and Makefile) are in the liveinstaller PR.

tests/images/azl-installer/iso/files/EULA.txt
alejandro-microsoft: It comes from the official Azure Linux installer. It was extracted from the [most recent ISO ](https://aka.ms/azurelinux-3.0-x86_64.iso). Let me verify if we need any specific approval to use it.

alejandro-microsoft: @Britel, I send you an email (End User License agreement) detailing the situation of the file. Do you think we should add the file as it is, or first talk with someone at CELA?

PR #273, opened 2025-10-16
bug: cloud-init disable fix

### LINKED ISSUE #272: Trident fails to disable cloud-init networking with cloud-init package is missing
Trident tries to disable cloud-init networking by invoking `disable_cloud_init_networking()`, but this function fail when the `cloud-init` package is missing in the image. This is because `disable_cloud_init_networking()` tries to write to `/etc/cloud/cloud.cfg.d` without first checking if the path exists or cloud-init is included in the image.

We should update Trident to confirm if `cloud-init` is included in the image first and also to check the configuration location path exists.

frhuelsz:
# 🔍 Description
 
Fix #272

Check if directory exists before attempting to create file.

### DIFF
--- a/crates/trident/src/subsystems/network.rs
+++ b/crates/trident/src/subsystems/network.rs
@@ -31,7 +31,7 @@ impl Subsystem for NetworkSubsystem {
                 // Trident is configuring the network, otherwise cloud-init may
                 // deploy additional configurations that are undesired and may
                 // conflict with or otherwise affect Trident's network setup.
-                disable_cloud_init_networking()?;
+                disable_cloud_init_networking(CLOUD_INIT_CONFIG_DIR)?;
             }
             None => {
                 debug!("Network config not provided");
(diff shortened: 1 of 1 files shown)


crates/trident/src/subsystems/network.rs
```diff
@@ -41,9 +42,22 @@ impl Subsystem for NetworkSubsystem {
 }
 
 fn disable_cloud_init_networking() -> Result<(), TridentError> {
-    fs::write(CLOUD_INIT_DISABLE_FILE, CLOUD_INIT_DISABLE_CONTENT)
+    let cloud_init_disable_path = Path::new(CLOUD_INIT_CONFIG_DIR).join(CLOUD_INIT_DISABLE_FILE);
+    if !Path::new(CLOUD_INIT_CONFIG_DIR).exists() {
```
jiria: I think this is sufficient to address this bug, but should we have a better handling re package presence detection in general? We could use the cosi manifest now, right?

crates/trident/src/subsystems/network.rs
```diff
@@ -41,9 +42,22 @@ impl Subsystem for NetworkSubsystem {
 }
 
 fn disable_cloud_init_networking() -> Result<(), TridentError> {
```
bfjelds: lets add a few UTs

frhuelsz: added

bfjelds (approved):

PR #283, opened 2025-10-23
engineering: Gate sysext and confext servicing on SELinux

ayaegashi:
# 🔍 Description

Gate sysext and confext servicing on SELinux not being in 'enforcing' mode. Also add static check. Both dynamic and static check mirror the way Trident currently handles root-verity and SELinux.

# 🤔 Rationale

The overlay mounted over /usr when merging sysexts causes the /usr directory to take on an unexpected label (should be usr_t but instead the directory is labeled tmpfs_t or user_tmpfs_t). This can cause other services to fail due to SELinux denials. Existing issue: https://github.com/systemd/systemd/issues/34387, patch not in v255 of systemd.

### DIFF
--- a/crates/trident/src/subsystems/selinux.rs
+++ b/crates/trident/src/subsystems/selinux.rs
@@ -513,7 +513,7 @@ mod tests {
         );
 
         // Test case 3: dynamic validation succeeds with SELinux in 'permissive'
-        // or 'disabled mode and root-verity present.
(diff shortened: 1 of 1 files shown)


bfjelds:
> Azure Pipelines successfully started running 1 pipeline(s).

i don't see a corresponding run to this message.  manually started a run.

PR #245, opened 2025-10-06
feature: uefi-fallback implementation

bfjelds:
# 🔍 Description
 
Implementing fallback for uefi variable corruption. Available modes are `none`, `rollback` and `rollforward`.
* `none` does not modify the UEFI fallback boot files.
* otherwise, a successful servicing will end with UEFI fallback configured to boot the target OS.

### `rollback`
| Trident Stage | `trident install` | `trident update` |
|---------------|-------------------|------------------|
| *stage* | UEFI fallback unchanged. | UEFI fallback unchanged. |
| *finalize* | UEFI fallback updated to boot target OS. | UEFI fallback updated to boot the servicing OS, meaning that any failures will cause the machine to boot into the servicing OS. |
| *commit* | No changes needed. | UEFI fallback updated to boot the target OS. |

### `rollforward`
| Trident Stage | `trident install` | `trident update` |
|---------------|-------------------|------------------|
| *stage* | UEFI fallback unchanged. | UEFI fallback unchanged. |
| *finalize* | UEFI fallback updated to boot target OS. | UEFI fallback updated to boot the target OS. |
| *commit* | No changes needed. | No changes needed. |

## Fallback Changes
* Add os.uefiFallback configuration with modes `rollback`, `rollforward`, and `none`
* For all e2e tests:
    * randomly add a uefifallback mode to the HC
    * add health-checks to validate the uefi fallback files before commit
    * add base e2e pytest to validate the uefi fallback files after commit
    * exception: validation not added to rerun because the HC needed for rerun + uefi fallback was too big to fit in ISO placeholder.  in future, if we care, we could make placeholder bigger.
* docs/Explanation files added for UEFI Variable and UEFI Fallback handling.

Validated: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=973916&view=results

PR #127, opened 2025-09-05
feature: create health-checks to enable user-defined validation that can trigger rollback

bfjelds:
# 🔍 Description
 
Enable customer-defined criteria that can trigger rollback without manual intervention.

Add health-checks that will trigger rollback for abupdate if they fail during commit-validate_boot flow.

## Rollback Changes
* Add health.checks section leveraging Scripts and creating SystemdCheck
    * health checks are run during validate_boot in commit
    * if failure occurs, they are saved in a file /var/lib/trident, enter HealthCheckFailed state, reboot is requested/invoked for ABUpdate
* Run autorollback scenario for all abupdate e2e tests
* Create e2e autorollback test for install
* Create rollback pytest

Validated: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=973917

PR #292, opened 2025-10-27
engineering: create abupdate scenario template

bfjelds:
# 🔍 Description

AbUpdate scenarios can be run for all e2e tests.

The scenario template uses storm to start an abupdate for the current configuration, passing any required parameters to storm. When storm finishes, a VM screenshot is grabbed. Finally, a pytest is run for the scenario for validation.

The following conditions must be met for the scenario to run
* abActiveVolume must not be 'null' ... this is the existing control mechanism for abupdates
* if the current test is a VM e2e tests, the scenario will run
* if the current test is a BM e2e test, the scenario will run in the weekly pipeline
* the current test configuration name must not be in the skipConfigurations parameter

Used by
* https://github.com/microsoft/trident/pull/127
* https://github.com/microsoft/trident/pull/245

Validated scenario template and conditions in the pipelines run for the above PRs:
* https://dev.azure.com/mariner-org/ECF/_build/results?buildId=965100&view=results
* https://dev.azure.com/mariner-org/ECF/_build/results?buildId=965187&view=results

### DIFF
--- a/.pipelines/templates/stages/testing_common/e2e-test-abupdate-scenario.yml
+++ b/.pipelines/templates/stages/testing_common/e2e-test-abupdate-scenario.yml
@@ -126,7 +126,7 @@ steps:
 
       PROXY_ARG=""
       if [ -n "${{ parameters.httpsProxy }}" ]; then
-        PROXY_ARG="--proxy HTTPS_PROXY=${{ parameters.httpsProxy }}"
+        PROXY_ARG="--env-vars HTTPS_PROXY=${{ parameters.httpsProxy }}"
       fi
 
       EXPECT_FAILURE_ARG=""


.pipelines/templates/stages/testing_common/e2e-test-abupdate-scenario.yml
```diff
@@ -0,0 +1,206 @@
...
+          --host ${{ parameters.hostIp }} \
+          --runtime-env ${{ parameters.runtimeEnv }} \
+          --configuration ${{ parameters.tridentConfigPath }} \
+          --ab-active-volume ${{ parameters.expectActiveVolume }}
+    timeoutInMinutes: 5
+    workingDirectory: $(Build.SourcesDirectory)/tests/e2e_tests
+    displayName: "🔬${{ parameters.updateEmoji }} Run Trident E2E tests after A/B update into target OS ${{ parameters.updateScenarioTargetOs }} testing ${{ parameters.updateScenarioName }}"
+    condition: and(succeeded(), ne(variables['abActiveVolume'], 'null'))
+
+  - bash: |
```
alejandro-microsoft: I'm a little confuse by this last part. Why do we always reset? Shouldn't we only reset for the rollback? Successful updates should maintain the change or what am I missing?

.pipelines/templates/stages/testing_common/display-serial-logs.yml
```diff
@@ -37,5 +41,5 @@ steps:
         echo "Serial log not found!"
         echo "(No serial log captured)" > $SERIAL_LOG_ARTIFACT_PATH
       fi
-    displayName: "📄 Display serial output from update to ${{ parameters.partition }}"
+    displayName: "📄 Display serial output from update to ${{ parameters.partition }} ${{ parameters.displayNameSuffix }}"
```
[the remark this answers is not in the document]
bfjelds: the serial log template is called multiple times during a pipeline, in the case of auto-rollback, the template will be called for the same partition 2x in a row: once for an update that auto-rollbacks, then for a subsequent update ... both targeting the same parition. this distinguishes the two.

.pipelines/templates/stages/testing_common/e2e-test-abupdate-scenario.yml
bfjelds: that's a good question.  for both rollback and uefifallback, it is appropriate.

but for some future scenario, it might not be right.

maybe it should be set to parameter.expectActiveVolume.

bfjelds: adjusted, thanks!

PR #293, opened 2025-10-27
engineering: enable --expect-failed-commit

bfjelds:
# 🔍 Description

For storm abupdate, allow validation if 'trident commit` fails expectedly. 

This requires a couple changes:

1. parameter and underlying code to judge a `trident commit` failure as expected
2. the service checking needs to survive when a reboot occurs, as it would when a health check fails

Used by
* https://github.com/microsoft/trident/pull/127
* https://github.com/microsoft/trident/pull/245

Validated expected failure and reboot resilience in the pipelines run for the above PRs:
* https://dev.azure.com/mariner-org/ECF/_build/results?buildId=965100&view=results
* https://dev.azure.com/mariner-org/ECF/_build/results?buildId=965187&view=results

### DIFF
--- a/tools/storm/helpers/check_ssh.go
+++ b/tools/storm/helpers/check_ssh.go
@@ -18,6 +18,7 @@ type CheckSshHelper struct {
 		utils.SshCliSettings `embed:""`
 		utils.EnvCliSettings `embed:""`
 		CheckActiveVolume    string `help:"Check that the indicated volume is the active one"`
+		ExpectFailedCommit   bool   `help:"Controls whether this test treats failed commits as successful." default:"false"`
 	}
 	client *ssh.Client
 }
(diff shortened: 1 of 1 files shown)


tools/storm/utils/trident.go
```diff
@@ -121,12 +116,15 @@ func CheckTridentService(client *ssh.Client, env TridentEnvironment, timeout tim
...
 
-	_, err := Retry(
+	reconnectNeeded, err := Retry(
 		timeout,
 		time.Second*5,
 		func(attempt int) (*bool, error) {
 			logrus.Infof("Checking Trident service status (attempt %d)", attempt)
-			err := checkTridentServiceInner(client, serviceName)
+			reconnect, err := checkTridentServiceInner(client, serviceName, expectSuccessfulCommit)
+			if reconnect != nil && *reconnect {
```
ayaegashi: If I'm understanding correctly, `reconnect` should never be nil and should always be True or False. Do we need to check that `reconnect != nil` ?

tools/storm/utils/trident.go
```diff
@@ -135,17 +138,20 @@ func CheckTridentService(client *ssh.Client, env TridentEnvironment, timeout tim
...
+	}
 	if err != nil {
 		return fmt.Errorf("trident service is not in expected state: %w", err)
 	}
 
 	return nil
 }
 
-func checkTridentServiceInner(client *ssh.Client, serviceName string) error {
+func checkTridentServiceInner(client *ssh.Client, serviceName string, expectSuccessfulCommit bool) (bool, error) {
```
frhuelsz: could you add a doc comment about the returns?

bfjelds: done

ayaegashi (approved):

PR #302, opened 2025-10-31
engineering: Add more UTs + FTs to pcrlock-related logic

ndubchak:
# 🔍 Description
 
This PR adds unit and functional testing to `engine/storage/encryption.rs`, as well as adds a validation step.

### DIFF
--- a/crates/trident/src/engine/storage/encryption.rs
+++ b/crates/trident/src/engine/storage/encryption.rs
@@ -532,10 +532,9 @@ mod tests {
 #[cfg(feature = "functional-test")]
 #[cfg_attr(not(test), allow(unused_imports, dead_code))]
 mod functional_test {
-    use pytest_gen::functional_test;
-
     use super::*;
 
+    use pytest_gen::functional_test;
     use trident_api::status::ServicingType;
 
     #[functional_test(feature = "helpers")]
--- a/crates/osutils/src/efibootmgr.rs
+++ b/crates/osutils/src/efibootmgr.rs
@@ -472,10 +472,11 @@ mod tests {
 #[cfg_attr(not(test), allow(unused_imports, dead_code))]
 mod functional_test {
     use super::*;
-    use pytest_gen::functional_test;
 
     use const_format::formatcp;
     use tempfile;
+
+    use pytest_gen::functional_test;
 
     use crate::testutils::repart::OS_DISK_DEVICE_PATH;
     use crate::{files::create_file, path::join_relative};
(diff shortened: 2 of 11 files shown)


crates/trident/src/engine/storage/encryption.rs
```diff
@@ -420,3 +435,174 @@ fn get_bootloader_paths(
...
+        ];
+        assert_eq!(
+            get_bootloader_paths(&esp_path, None, &ctx).unwrap(),
+            expected_paths_b
+        );
+
+        // Test case #4: If active volume is None, but servicing type is not clean install and
+        // mount_path is provided, return an error.
+        let mount_path = PathBuf::from("/mnt");
+        ctx.servicing_type = ServicingType::AbUpdate;
```
ayaegashi: Wondering, is this necessary? It doesn't look like `get_bootloader_paths` checks servicing type

ndubchak: Yes, the function doesn't explicitly check the servicing type, but I thought it was useful for the understanding (i.e. `mount_path` is only relevant on encryption provisioning, which is never executed on clean install) if the tests need to be extended/corrected/re-factored in the future. But if this is confusing on the contrary, will remove

ayaegashi: I don't feel strongly about it, but perhaps a short comment would be useful if you want to keep it!

crates/trident/src/engine/storage/encryption.rs
```diff
@@ -319,6 +319,22 @@ pub fn get_binary_paths_pcrlock(
         vec![]
     };
 
+    // Validate that these paths exist to fail early
```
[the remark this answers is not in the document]
ndubchak: Hm, I mean the following here: "Validate that these paths exist here, i.e. before we actually manipulate them, so that we can fail early"

crates/trident/src/engine/storage/encryption.rs
```diff
@@ -281,8 +281,8 @@ fn encrypt_and_open_device(
 }
 
 /// Returns paths of UKI and bootloader binaries that `systemd-pcrlock` tool should seal to. During
-/// encryption provisioning, returns binaries used for the current boot, as well as binaries that
-/// will be used in the future boot, i.e. in the target OS image. During rollback validation,
+/// encryption provisioning, returns binaries used for the current boot, , i.e. servicing OS, as
```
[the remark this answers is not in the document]
ndubchak: Corrected

crates/trident/src/engine/storage/encryption.rs
ndubchak: Actually, sorry, we do need to set the servicing type here. The logic called by `boot::make_esp_dir_name()` requires it for constructing the path, so keeping it

ayaegashi (approved):

PR #303, opened 2025-10-31
engineering: isopatcher tool

alejandro-microsoft:
# 🔍 Description
 
This PR relocates the function `PatchFile`, which patches an ISO with a specific file (it uses a placeholder to inject files into the ISO without rebuilding it). The function is moved from `netlaunch` tool to a common package that can be reused by more tools. 

This PR also introduces the `isopatch` tool, a general-purpose utility for patching ISO images with custom files. 

# 🤔 Rationale

We can use this tool to change an installer ISO from attended installation to unattended installation by inserting a Host Configuration.


bfjelds (changes requested):
revert oras change if not needed

PR #304, opened 2025-10-31
 engineering: Installer ISO definition

alejandro-microsoft:
# 🔍 Description
 
- ISO definition to produce the ISO for attended installation by default, with possibility to patch the ISO with a Host Configuration for unattended installation.
- Disabled the EULA view and requirement (this can be re-enabled when needed).
- Modifications to paths in builder to use it for building the installer ISO.
- Make rule to generate the ISO.
- Moved trident-mos to the images directory (**only moved, no changes except for the paths change on `iso.yaml`**) to be grouped with the other images.

### DIFF
--- a/tests/images/trident-mos/iso.yaml
+++ b/tests/images/trident-mos/iso.yaml
@@ -63,7 +63,7 @@ os:
     # Required for systemd-pcrlock & PCR-based encryption; can be removed once
     # AZL 4.0 merges a fix to provide statically defined files inside the same
     # package as the systemd-pcrlock binary, i.e. systemd-udev.
-    - source: "../../packaging/static-pcrlock-files/"
+    - source: "../../../packaging/static-pcrlock-files/"
       destination: "/var/lib/pcrlock.d/"
 
   additionalFiles:
@@ -73,15 +73,15 @@ os:
       destination: /usr/lib/systemd/system/serial-getty@.service
     - source: files/root.profile
       destination: /root/.profile
-    - source: ../../packaging/systemd/trident-install.service
+    - source: ../../../packaging/systemd/trident-install.service
       destination: /usr/lib/systemd/system/trident-install.service
     - source: files/download-trident.service
       destination: /usr/lib/systemd/system/download-trident.service
-    - source: ../../packaging/selinux-policy-trident/trident.if
+    - source: ../../../packaging/selinux-policy-trident/trident.if
       destination: /usr/share/selinux/packages/trident/trident.if
-    - source: ../../packaging/selinux-policy-trident/trident.fc
+    - source: ../../../packaging/selinux-policy-trident/trident.fc
       destination: /usr/share/selinux/packages/trident/trident.fc
-    - source: ../../packaging/selinux-policy-trident/trident.te
+    - source: ../../../packaging/selinux-policy-trident/trident.te
       destination: /usr/share/selinux/packages/trident/trident.te
 
   services:


bfjelds (dismissed):

Makefile
```diff
@@ -430,33 +430,45 @@ bin/virtdeploy: tools/cmd/virtdeploy/* tools/go.sum tools/pkg/* tools/pkg/virtde
...
-# If necessary create End-User License Agreement example file in execution directory
-bin/EULA.txt:
-	@mkdir -p bin
-	@echo "SAMPLE EULA" > $@
-
-# EULA.txt required at runtime; added to ensure binary will be able to run
 bin/liveinstaller: \
 	$(shell find $(INSTALLER_DIR)/ -type f) \
-	$(INSTALLER_DIR)/go.sum \
-	bin/EULA.txt
```
ndubchak: What was `bin/EULA.txt` here responsible for?

Makefile
```diff
@@ -430,33 +430,45 @@ bin/virtdeploy: tools/cmd/virtdeploy/* tools/go.sum tools/pkg/* tools/pkg/virtde
...
+	$(INSTALLER_DIR)/go.sum
 	@mkdir -p bin
 	cd $(INSTALLER_DIR)/imagegen/attendedinstaller/attendedinstaller_tests && \
 		CGO_ENABLED=0 go build -o $(CURDIR)/$(INSTALLER_OUT_DIR)/attendedinstaller-simulator attendedinstaller_simulator.go
 
 .PHONY: run-attendedinstaller-simulator
-run-attendedinstaller-simulator: bin/attendedinstaller-simulator bin/EULA.txt
+run-attendedinstaller-simulator: bin/attendedinstaller-simulator
```
ndubchak: kind of a silly Q, but why is this `run-attendedinstaller-simulator` and not `run-attended-installer-simulator`?

Makefile
```diff
@@ -430,33 +430,45 @@ bin/virtdeploy: tools/cmd/virtdeploy/* tools/go.sum tools/pkg/* tools/pkg/virtde
...
+artifacts/test-image/azl-installer.iso: \
+	bin/RPMS \
+	bin/liveinstaller \
+	artifacts/test-image/regular.cosi \
+	tests/images/azl-installer/installer-iso.yaml \
+	$(shell find tests/images/azl-installer/ -type f 2>/dev/null)
+	# Copy runtime image
+	rm -rf tests/images/azl-installer/iso/images
+	mkdir -p tests/images/azl-installer/iso/images
+	cp artifacts/test-image/regular.cosi tests/images/azl-installer/iso/images/trident-testimage.cosi
```
ndubchak: nit: Seems like a few of these could be declared as constants so that we don't rewrite the same paths multiple times; e.g. `artifacts/test-image` and `tests/images/azl-installer`

Makefile
```diff
@@ -388,7 +388,7 @@ generate-functional-test-manifest: .cargo/config
 
 .PHONY: validate-configs
 validate-configs: bin/trident
-	$(eval DETECTED_HC_FILES := $(shell grep -R '^storage:' . --include '*.yaml' -l | grep -E -v '\./(tests/trident-mos|target|dev|azure-linux-image-tools|crates/docbuilder|tests/images)'))
+	$(eval DETECTED_HC_FILES := $(shell grep -R '^storage:' . --include '*.yaml' -l | grep -E -v '\./(target|dev|azure-linux-image-tools|crates/docbuilder|tests/images)'))
```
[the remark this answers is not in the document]
alejandro-microsoft: I was thinking the same thing, but when I removed it, the check failed. I was going to dig a more into it, but I think it would be netter to do it in a follow-up PR.

Makefile
alejandro-microsoft: It is needed to show the screen to accept the End-User License Agreement. The rule created a dummy file, if there wasn't a document provided. We decided to disable the screen until needed, so it is no longer necessary.

tests/images/azl-installer/iso/files/config-placeholder.yaml
```diff
@@ -0,0 +1 @@
...################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################################
```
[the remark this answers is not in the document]
alejandro-microsoft: To give the option to patch the ISO by adding a Host Configuration. Adding it would change the ISO from attended installation to unattended installation. That way we have an easy way to have an attended installer by just patching the ISO.

Makefile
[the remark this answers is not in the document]
alejandro-microsoft: I look into it, and the problem is that when we build the dev version of osmodifier, we clone and use `azure-linux-image-tools`. The stage that runs the makefile validation, also builds the dev version of the osmodifier: https://github.com/microsoft/trident/blob/main/.pipelines/templates/stages/validate_makefile/dev-build.yml#L37. So, we still need to exclude it.

ndubchak (approved):

bfjelds (approved):

PR #313, opened 2025-11-06
refactor validate_boot

bfjelds:
# 🔍 Description
 
`validate_boot` is large and kind of a mixed bag of functionality. Attempt to clarify with refactor.

### DIFF
--- a/crates/trident/src/engine/rollback.rs
+++ b/crates/trident/src/engine/rollback.rs
@@ -56,8 +56,9 @@ pub fn commit(datastore: &mut DataStore) -> Result<BootValidationResult, Trident
             return Err(err);
         }
     }
-    // If we reached here, we booted from the expected root device. Persist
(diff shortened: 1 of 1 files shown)


bfjelds:
@copilot

ndubchak:
Other possibilities:

1. Rename `rollback.rs` to `commit.rs` and reflect that we're not just handling rollback
2. Implement this whole commit logic as a whole separate stage in the servicing:

- commit is itself a subsystem step like `provision()` or `configure()`
- commit is at the level of stage/finalize; its methods can be implemented as methods of Subsystems, similarly to `provision()` or `configure()`
- commit as a separate action like install/update, which would make sense because we have a separate `trident commit`; it could have its own stage/finalize; since updating the boot order, re-generating pcrlock policy, etc. is finalizing all this stuff

PR #322, opened 2025-11-11
feature: Add support for 'trident stream-image' command

fintelia:
This adds a command for trident to install based only on the URL of a COSI file and optionally its hash. The HC itself is embedded inside the COSI file and the `tera` template engine is used to enable specialization of the HC file based on the system (intended mostly to select which disk to install onto).

Currently gated behind the `dangerous-options` feature so it won't be exposed in release builds.

### DIFF
--- a/crates/trident/src/stream.rs
+++ b/crates/trident/src/stream.rs
@@ -62,7 +62,9 @@ pub fn config_from_image_url(
 ///
 /// # Provided Context
 ///
-/// * `disks`: A list of detected block devices of type `disk`. Each disk has the following fields:
+/// * `get_disks()`: Returns a list of detected block devices of type `disk`. Accepts optional
+///   arguments `min_size` and `max_size` to filter disks by size in bytes. Each disk has the
+///   following fields:
 ///   * `name`: The device name (e.g., `sda`, `nvme0n1`).
 ///   * `path`: The full device path (e.g., `/dev/sda`, `/dev/nvme0n1`).
 ///   * `size`: The size of the device in bytes.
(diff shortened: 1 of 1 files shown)


crates/trident/src/osimage/cosi/metadata.rs
```diff
@@ -297,6 +301,13 @@ pub(crate) struct BootloaderEntry {
     pub cmdline: String,
 }
 
+#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub(crate) struct AuxiliaryFile {
```
ndubchak: Do we want to name this struct in a generic way for later re-use? Or should the naming be specific to the HC template?

crates/trident/src/stream.rs
```diff
@@ -0,0 +1,220 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_expand_template_basic_context() {
+        assert_eq!(
+            expand_template(
+                r#"
```
frhuelsz: nit: indoc for more agreeable indenting

crates/trident/src/stream.rs
```diff
@@ -0,0 +1,220 @@
...
+"#
+        )
+        .is_ok());
+    }
+
+    #[test]
+    fn test_expand_template_size_range_filter() {
+        let result = expand_template(
+            r#"
+small_disks: {{ disks | size_range(high=1073741824) }}
```
frhuelsz: nit: this syntax may be more confusing, i'd suggest just having a function called `disk_selector` or similar which will read the disks internally, instead of using a value and a filter

crates/trident/src/stream.rs
```diff
@@ -0,0 +1,220 @@
...
+    impl tera::Filter for SizeRange {
+        fn filter(
+            &self,
+            value: &tera::Value,
+            args: &HashMap<String, tera::Value>,
+        ) -> tera::Result<tera::Value> {
+            match value.clone() {
+                tera::Value::Array(a) => {
+                    let low = match args.get("low") {
+                        Some(v) => v.as_u64().unwrap_or(0),
```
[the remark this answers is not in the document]
fintelia: tera supports arithmetic so you can pass something like `3*GB` to the filter

crates/trident/src/stream.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    config.image = Some(config::OsImage {
+        url: image_url,
+        sha384: image_source.sha384,
+    });
+
+    Ok(config)
+}
+
+fn expand_template(template: &str) -> Result<String, anyhow::Error> {
```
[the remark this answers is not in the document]
fintelia: Not sure the right approach here. The challenge of template expansion is that it might not produce the same results if run multiple times.

bfjelds (approved):

PR #330, opened 2025-11-12
WIP: engineering: allow rerun of update that has failed health check

bfjelds:
# 🔍 Description
 
Trident currently has special casing to allow rerun of `trident udpate` for failed update in AbUpdateFinalized state. Enable similar for AbUpdateHealthCheckFailed state.


bfjelds:
> Do we need to then update the state diagram(s)?

i don't think so.  but i'm going to do a little more thinking on states before merging this.

PR #340, opened 2025-11-17
WIP: add AbUpdateRollbackFailed state, add state machine doc with updated content from devdocs

bfjelds:
# 🔍 Description
 
Try to avoid dead-end states. Create AbUpdateRollbackFailed. Create public state machine docs.

e2e-pr: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=984740&view=results

### DIFF
--- a/docs/Explanation/Trident-States.md
+++ b/docs/Explanation/Trident-States.md
@@ -127,24 +127,43 @@ When troubleshooting Trident servicing issues, it is important to check both
 the `ServicingState` and `LastError` of the host in the Host Status. The
 Host Status can be viewed using the `trident get status` command.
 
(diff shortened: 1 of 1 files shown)


crates/trident_api/src/status.rs
```diff
@@ -99,6 +99,9 @@ pub enum ServicingState {
     /// A/B update has been finalized. For the next boot, the firmware will boot from the updated
     /// target OS image.
     AbUpdateFinalized,
+    /// A/B update has failed, and Trident initiated an auto-rollback, but the host failed to
+    /// successfully rollback to the servicing OS.
+    AbUpdateRollbackFailed,
```
frhuelsz: note: this will become an official API name, it should be reviewed/approved

PR #344, opened 2025-11-19
engineering: share e2e screenshot code with servicing

bfjelds:
# 🔍 Description
 
VM screenshots are captured currently in the e2e tests using a storm scriptset.  The code is not shareable with other helpers/scenarios/scriptsets.  This PR moves the implementation to a shareable location and adds screenshot capture to the servicing storm scenario.

### DIFF
--- a/tools/storm/utils/libvirt.go
+++ b/tools/storm/utils/libvirt.go
@@ -158,17 +158,17 @@ func CaptureScreenshot(vmName string, artifactsFolder string, screenshotFilename
 
 	err = capturePpmScreenshot(vmName, ppmFilename.Name())
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create PPM screenshot: %w", err)
 	}
 
 	err = os.MkdirAll(artifactsFolder, 0755)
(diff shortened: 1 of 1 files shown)


tools/storm/scripts/capture_screenshot/capture_screenshot.go
```diff
@@ -20,44 +17,9 @@ type CaptureScreenshotScript struct {
...
-
-	pngPath := filepath.Join(s.ArtifactsFolder, s.ScreenshotFilename)
-	if err := convertPpmToPng(ppmFilename.Name(), pngPath); err != nil {
-		return err
-	}
-	return nil
-}
-
-func capturePpmScreenshot(vmName string, ppmFilename string) error {
-	virshOutput, virshErr := exec.Command("sudo", "virsh", "screenshot", vmName, ppmFilename).CombinedOutput()
```
jiria: >virshOutput [](http://example.com/codeflow?start=1&length=11)

Would it make sense to put some info into the description as to what is wrong with the current method?

PR #350, opened 2025-11-21
bug: Fix extension-release file name

ayaegashi:
# 🔍 Description
 
Extension-release file is not being named correctly by the build_extension_images.go script, because the name of the image file should match the suffix of the extension-release file.


tools/storm/scripts/build_extension_images/build_extension_images.go
```diff
@@ -58,6 +58,7 @@ func (s *BuildExtensionImagesScript) Run() error {
 
 func buildImage(extType string, numClones int) error {
 	for i := 1; i <= numClones; i++ {
+		extName := fmt.Sprintf("test-%s-%d", extType, i)
```
bfjelds: does this mean that the extensions are no longer versions of the same extension?

PR #371, opened 2025-12-03
engineering: env::set_var and env::remove_var unsafe in rust 2024, fix clippy warnings

ayaegashi:
# 🔍 Description
 
`env::set_var` and `env::remove_var` were marked unsafe in Rust 2024 edition (https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html). Trident only uses these functions in tests and in docbuilder. To fix clippy warnings, this PR adds the `unsafe {}` blocks where we use these functions.

### DIFF
--- a/crates/osutils/src/container.rs
+++ b/crates/osutils/src/container.rs
@@ -208,6 +208,11 @@ mod functional_test {
             std::fs::create_dir(test_dir).unwrap();
         }
         assert_eq!(get_host_root_path().unwrap(), Path::new(HOST_ROOT_PATH));
+
+        // Clean-up
+        unsafe {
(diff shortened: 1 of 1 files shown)


frhuelsz (approved):

ayaegashi:
> food for thought: the use of env vars like that is... risky and only works because we don't run functional tests in parallel, but in the future we may want to look into something like https://github.com/vmx/temp-env to make things less brittle

I was also looking into `serial_test` (https://docs.rs/serial_test/latest/serial_test/) which prevents tagged tests from running in parallel, although as you said this isn't necessary with functional tests.

PR #386, opened 2025-12-09
engineering: Harpoon gRPC Protobuf

frhuelsz:
# 🔍 Description
 
Introduce Harpoon gRPC proto file

### DIFF
--- a/crates/harpoon/Cargo.toml
+++ b/crates/harpoon/Cargo.toml
@@ -2,6 +2,7 @@
 name = "harpoon"
 version = "0.1.0"
 edition = "2024"
+license = "MIT"
 
 [dependencies]
 prost = "0.14.1"


proto/harpoon.proto
```diff
@@ -0,0 +1,229 @@
...
+
+    // RebuildRaid performs a RAID rebuild operation to restore a degraded RAID
+    // array on new disks.
+    rpc RebuildRaid(RebuildRaidRequest) returns (stream ServicingResponse);
+
+    // Validates a Host Configuration file.
+    rpc ValidateHostConfig(ValidateRequest) returns (ValidateResponse);
+
+    // Returns the servicing type required to apply the given Host Configuration.
+    rpc GetServicingType(ServicingTypeRequest) returns (ServicingTypeResponse);
```
ayaegashi: A bit out of scope for this PR maybe, but are you thinking this would correspond to a new verb in the Trident API as well?

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
+syntax = "proto3";
+
+package harpoon.v1;
+
+import "google/protobuf/timestamp.proto";
+
+// TridentService provides methods for managing Trident servicing operations
+service TridentService {
+  // Install performs a Trident installation operation.
```
ndubchak: This is kind of a nit, but do we still use "operation" to refer to the allowed operations? Because then I think we should re-word to:

"Install performs a Trident installation servicing." or "Install performs a Trident clean install."

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+  rpc Install(ServicingRequest) returns (stream ServicingResponse);
+
+  // InstallStage perform the stage part of a Trident installation operation.
+  rpc InstallStage(StagingRequest) returns (stream ServicingResponse);
```
ndubchak: another nit: Should we either say `StagingRequest` and `FinalizingRequest` or `StageRequest` and `FinalizeRequest`?

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+
+  // Stream image to the target device.
+  rpc StreamImage(StreamImageRequest) returns (stream ServicingResponse);
+
+  // RebuildRaid performs a RAID rebuild operation to restore a degraded RAID
+  // array on new disks.
+  rpc RebuildRaid(RebuildRaidRequest) returns (stream ServicingResponse);
+
+  // Validates a Host Configuration file.
+  rpc ValidateHostConfig(ValidateRequest) returns (ValidateResponse);
```
ndubchak: Are we planning to have any other operations or terms with "validation"? Because if yes, should this be a `ValidateHostConfigRequest`?

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+  rpc UpdateStage(StagingRequest) returns (stream ServicingResponse);
+
+  // UpdateFinalize performs the finalize part of a Trident update operation.
+  rpc UpdateFinalize(FinalizeRequest) returns (stream ServicingResponse);
+
+  // Check root validates that the host booted using the expected root
+  // filesystem.
+  rpc CheckRoot(CheckRootRequest) returns (stream ServicingResponse);
+
+  // Attempts to commit a finalized update.
```
ndubchak: Is there no `CommitInstall`?

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+
+  // RebuildRaid performs a RAID rebuild operation to restore a degraded RAID
+  // array on new disks.
+  rpc RebuildRaid(RebuildRaidRequest) returns (stream ServicingResponse);
+
+  // Validates a Host Configuration file.
+  rpc ValidateHostConfig(ValidateRequest) returns (ValidateResponse);
+
+  // Returns the servicing type required to apply the given Host Configuration.
+  rpc GetServicingType(ServicingTypeRequest) returns (ServicingTypeResponse);
```
ndubchak: Do we maybe need both `GetRequiredServicingType` and `GetServicingType`? Also, `GetServicingTypeRequest` to match the other types where "get" is included?

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+  FileLocation location = 5;
+}
+
+// Represents the final status of a servicing operation.
+enum StatusCode {
+  SUCCESS = 0;
+  FAILURE = 1;
+}
+
+message FinalStatus {
```
ndubchak: Would this essentially be the Host Status? Or, also some other things?

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+message GetLastErrorResponse {
+  // The last error encountered during servicing.
+  optional TridentError error = 1;
+}
+
+message GetServicingStateRequest {}
+
+// Mapped to existing Trident servicing states
+enum ServicingState {
+  NOT_PROVISIONED   = 0;
```
ndubchak: Do we want to add the `RuntimeX` states? Also `AbUpdateHealthCheckFailed`

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+  A = 0;
+  B = 1;
+}
+
+message GetActiveVolumeResponse {
+  ABVolumeState active_volume = 1;
+}
+
+message ValidateRequest {
+  // The configuration to be validated.
```
ndubchak: nit: Host Configuration

Or is this for different types of configs?

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+
+message ValidateResponse {
+  // Indicates if the configuration is valid.
+  bool            valid   = 1;
+  // Message providing details about validation errors.
+  optional string message = 2;
+}
+
+enum ServicingType {
+  // No servicing is currently in progress nor required.
```
ndubchak: Wait, so is this the current servicing type or the one required by the HC?

proto/harpoon.proto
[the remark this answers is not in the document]
frhuelsz: I think install is a more generic name, and "clean install" is a specific subset of installs that was used because we only supported clean installs at the beginning. I'd rather not perpetuate the pattern in new apis, this is also shorter.

proto/harpoon.proto
frhuelsz: updated for consistency

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+  rpc StreamImage(StreamImageRequest) returns (stream ServicingResponse);
+
+  // RebuildRaid performs a RAID rebuild operation to restore a degraded RAID
+  // array on new disks.
+  rpc RebuildRaid(RebuildRaidRequest) returns (stream ServicingResponse);
+
+  // Validates a Host Configuration file.
+  rpc ValidateHostConfig(ValidateRequest) returns (ValidateResponse);
+
+  // Returns the servicing type required to apply the given Host Configuration.
```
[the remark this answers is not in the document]
frhuelsz: clarified a bit

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+  rpc GetActiveVolume(GetActiveVolumeRequest) returns (GetActiveVolumeResponse);
+}
+
+message StagingRequest {
+  // The target Host Configuration.
+  string config = 1;
+}
+
+message FinalizeRequest {
+  // Indicates if the orchestrator handles the reboot process.
```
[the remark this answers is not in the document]
frhuelsz: yes, for some scenarios an orchestrator may need to do some cleanup in between finalize and a reboot, in such cases the orchestrator would say `true` here and upcon completion, expect Trident to say whether a reboot is needed or no, and do it when necessary.

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+// Represents the final status of a servicing operation.
+enum StatusCode {
+  SUCCESS = 0;
+  FAILURE = 1;
+}
+
+message FinalStatus {
+  // Final status of the servicing operation.
+  StatusCode status          = 1;
+  // Message associated with the final status.
```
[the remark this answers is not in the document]
frhuelsz: tbh this we may need to revise as we go

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+  // Validates a Host Configuration file.
+  rpc ValidateHostConfig(ValidateRequest) returns (ValidateResponse);
+
+  // Returns the servicing type required to apply the given Host Configuration.
+  rpc GetServicingType(ServicingTypeRequest) returns (ServicingTypeResponse);
+
+  // Get the last provisioned configuration, if any.
+  rpc GetCurrentConfig(GetConfigRequest) returns (GetConfigResponse);
+
+  // Get the active servicing configuration, if any.
```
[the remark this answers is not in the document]
frhuelsz: correct, servicing config would only return something after stage and before commit

proto/harpoon.proto
```diff
@@ -0,0 +1,238 @@
...
+
+message GetServicingStateResponse {
+  // The current servicing state of Trident.
+  ServicingState state = 1;
+}
+
+message GetActiveVolumeRequest {}
+
+// Represents the active volume in a dual-volume setup.
+enum ABVolumeState {
```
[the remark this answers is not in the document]
ndubchak: So, are we resolving this?

ndubchak (dismissed):

proto/harpoon.proto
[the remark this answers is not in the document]
frhuelsz: For the time being, yes, Trident does have a valid None state right now and we need to capture that. Discussing whether that should be the case at all is more of a design question that's a bit outside of the scope of this PR.

ndubchak (approved):

PR #387, opened 2025-12-09
bug: Check extension image path in HC; remove incompatible confext testing

ayaegashi:
# 🔍 Description
 
This PR fixes a couple of bugs in the extensions testing flow:
1. Trident does not currently check the paths of sysexts and confexts in the Host Configuration. This means that in the target OS they will not properly merge. This PR adds this validation.
2. This was slipping by in tests because `systemd-sysext list` outputs all extension-image like objects on the OS, without checking if they are valid or active. `systemd-sysext status` only returns actively merged extensions, which is what we want.
3. Confexts were not being successfully merged. While Trident could add it to the OS in the `root-verity` test, because there is already a `trident-overlay` mounted at `/etc`, when `systemd-confext` attempts to mount another overlay at `/etc` the merging fails. While Trident succeeds, the enabled `systemd-confext` service is not succeeding. This PR removes testing for confexts and updates the confext documentation.

### DIFF
--- a/crates/trident_api/schemas/host-config-schema.json
+++ b/crates/trident_api/schemas/host-config-schema.json
@@ -327,7 +327,7 @@
       ],
       "properties": {
         "path": {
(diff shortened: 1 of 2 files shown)


docs/Explanation/Confexts.md
```diff
@@ -44,6 +44,10 @@ are not currently supported in Azure Linux 3.0 (systemd v255). It is important
 to note that configuring confexts will result in `/etc` becoming read-only. This
 can be problematic if anything requires writing to `/etc`.
 
+Note: confext servicing is incompatible with root-verity since Trident's
```
bfjelds: Does the extension need to be .raw?  If so, should that be mentioned in doc?

ayaegashi: Good point, will add.

PR #390, opened 2025-12-10
engineering: Handle PCR encryption for manual rollback

ndubchak:
# 🔍 Description
 
This PR handles the PCR encryption for the manual rollback scenario. 

In particular, when staging a manual rollback that requires a reboot, with a UKI image + encryption, Trident will re-generate the pcrlock policy to include both the current and the future (rollback) boots.

### DIFF
--- a/crates/trident/src/engine/storage/encryption.rs
+++ b/crates/trident/src/engine/storage/encryption.rs
@@ -367,7 +367,7 @@ fn get_uki_paths(
         // executed while staging an A/B update
         if manual_rollback {
             return Err(anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


crates/trident/src/engine/rollback.rs
```diff
@@ -265,6 +265,37 @@ fn commit_finalized_on_expected_root(
 
     // If this is a UKI image, then we need to re-generate pcrlock policy to include the PCRs
     // selected by the user for the current boot only.
+    // If it's virtdeploy, after confirming that we have booted into the correct image, we need
```
ndubchak: TODO: Remove b/c bad rebase

PR #394, opened 2025-12-11
engineering: Improve COSI Metadata Parsing

frhuelsz:
# 🔍 Description
 
- Remove deny_unknown_fields for forwards compatibility
- Introduce an enum of known cosi versions to ensure we keep track of what is expected of each
- Parse spec enums as rust enums.

### DIFF
--- a/crates/trident/src/osimage/cosi/validation.rs
+++ b/crates/trident/src/osimage/cosi/validation.rs
@@ -71,7 +71,7 @@ impl CosiMetadata {
 
                 // Unknown bootloader type is warned about, it may cause issues down the line
                 (BootloaderType::Unknown(bootloader_type), _) => {
-                    log::warn!("Unknown bootloader type: {}", bootloader_type)
+                    warn!("Unknown bootloader type: {}", bootloader_type)
                 }
 
                 // Everything else is OK


crates/trident/src/osimage/cosi/metadata.rs
```diff
@@ -82,56 +104,66 @@ impl CosiMetadata {
...
-                if self.version.major > 1 || (self.version.major == 1 && self.version.minor > 0) {
-                    bail!("Bootloader is required for COSI version >= 1.1, but not provided");
+
+                // Unknown bootloader type is invalid
+                (BootloaderType::Unknown(bootloader_type), _) => {
+                    bail!("Unsupported bootloader type: {}", bootloader_type);
                 }
+
+                // Valid combination
+                _ => {}
```
ndubchak: What is this for?

frhuelsz: this is the wildcard for all non-error conditions, this whole  statement is only trying to match on error conditions

crates/trident/src/osimage/cosi/metadata.rs
```diff
@@ -82,56 +104,66 @@ impl CosiMetadata {
...
-                    _ => bail!("Unsupported bootloader type: {}", bootloader_type),
                 }
-            }
-            None => {
-                if self.version.major > 1 || (self.version.major == 1 && self.version.minor > 0) {
-                    bail!("Bootloader is required for COSI version >= 1.1, but not provided");
+
+                // Unknown bootloader type is invalid
+                (BootloaderType::Unknown(bootloader_type), _) => {
+                    bail!("Unsupported bootloader type: {}", bootloader_type);
```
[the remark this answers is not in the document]
frhuelsz: This is more about this version of trident not recognizing this bootloader type. If we see something we don't recognize then we have no idea what to do with it?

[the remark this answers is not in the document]
fintelia: I'd feel better about running this once we know the specific operation we're doing with the COSI. Perhaps directly in the validate() method of the subsystem that is using the bootloader field?

For something like `trident stream-image` it might not matter and would present a forward compatibility concern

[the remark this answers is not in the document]
frhuelsz: Fair. This is the existing logic that was implemented before, but happy to revise for forward compat.

crates/trident/src/osimage/cosi/metadata.rs
```diff
@@ -82,56 +104,74 @@ impl CosiMetadata {
...
+                    bail!("Bootloader type 'systemd-boot' requires systemd-boot entries");
+                }
+
+                // Systemd-boot with not exactly 1 UKI entry is invalid for this version of Trident
+                (BootloaderType::SystemdBoot, Some(systemd_boot)) => {
+                    match systemd_boot.entries.as_slice() {
+                        // No entries is invalid
+                        [] => bail!("Bootloader type 'systemd-boot' must not be empty"),
+
+                        // First entry MUST be of type 'uki-standalone'
```
ayaegashi: Why do we only warn if the first entry _must_ be 'uki-standalone'? Shouldn't we error out?

crates/trident/src/osimage/cosi/metadata.rs
```diff
@@ -82,56 +104,74 @@ impl CosiMetadata {
...
+                        [entry, ..] if !entry.boot_type.eq(&SystemdBootloaderType::UkiStandalone) => log::warn!(
+                            "First entry of bootloader type 'systemd-boot' is not of type 'uki-standalone'"
+                        ),
+
+                        // Having more than one entry is warned about, only the first is used in this version of Trident.
```
ayaegashi: Similar question, why don't we return an error here if its considered invalid?

crates/trident/src/osimage/cosi/metadata.rs
```diff
@@ -301,6 +368,25 @@ pub(crate) struct BootloaderEntry {
     pub cmdline: String,
 }
 
+#[derive(Debug, Deserialize, Clone, Eq, PartialEq, Display)]
+pub(crate) enum SystemdBootloaderType {
+    #[serde(rename = "uki-standalone")]
```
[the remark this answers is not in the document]
frhuelsz: probably yes, but i went for an overly cautions approach to ensure we have the exact same enum definitions as the spec

crates/trident/src/osimage/cosi/metadata.rs
frhuelsz: We would eventually fail, but yes, we could use more validation, will create an issue to tackle later. The immediate goal is to relax most of the checking here as that has caused us issues before.

ayaegashi (approved):

PR #413, opened 2025-12-16
infra: add proper context to netlaunch

frhuelsz:
# 🔍 Description
 
Add a `context.Context` parameter to the netlaunch core function so that it may be externally cancelled.

### DIFF
--- a/tools/pkg/netlaunch/netlaunch.go
+++ b/tools/pkg/netlaunch/netlaunch.go
@@ -22,7 +22,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-func RunNetlaunch(config *NetLaunchConfig) error {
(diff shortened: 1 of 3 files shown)


tools/pkg/netlaunch/netlaunch.go
```diff
@@ -258,7 +259,7 @@ func RunNetlaunch(config *NetLaunchConfig) error {
 	// Wait for something to happen
 	exitError := phonehome.ListenLoop(terminateCtx, result, config.WaitForProvisioning, config.MaxPhonehomeFailures)
 
-	err = server.Shutdown(context.Background())
+	err = server.Shutdown(ctx)
```
ndubchak: Why are we using `ctx` here as opposed to `bmcCtx`?

ayaegashi (approved):

PR #421, opened 2025-12-18
bug: Pass Netplan previous state

ayaegashi:
# 🔍 Description
 
Netplan operates statelessly so it will not remove virtual devices that no longer exist in a Netplan config (see: https://netplan.readthedocs.io/en/stable/netplan-apply/#known-issues). We can get around this by passing the previous Netplan configuration to Netplan with `--state`.

Note: this PR backs up for the config currently at Trident's Netplan config path, not the Netplan config in the old HC.

### DIFF
--- a/crates/osutils/src/netplan.rs
+++ b/crates/osutils/src/netplan.rs
@@ -76,7 +76,7 @@ pub fn backup() -> Result<(), Error> {
     Ok(())
 }
 
-/// Clean up netplan back-up directory.
+/// Clean up netplan backup directory.
 pub fn cleanup_backup() -> Result<(), Error> {
     if fs::exists(NETPLAN_BACKUP_DIR)? {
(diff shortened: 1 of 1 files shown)


crates/osutils/src/netplan.rs
```diff
@@ -42,6 +53,37 @@ fn render_netplan_yaml(value: &NetworkConfig) -> Result<String, Error> {
...
+    // Re-create the directory.
+    if let Some(parent) = Path::new(NETPLAN_BACKUP_FILE).parent() {
+        fs::create_dir_all(parent).with_context(|| {
+            format!(
+                "Failed to create netplan config backup directory at '{}'",
+                parent.display()
+            )
+        })?;
+    }
+    // Copy over existing netplan config if it exists.
```
bfjelds: i'm curious what happens if there is another netplan file in the netplan folder (if a customer creates one with additionalFiles or something).

PR #553, opened 2026-03-05
engineering: start using patch in trident version for pipelines

bfjelds:
# 🔍 Description
 
Change versioning to include patch: `0.22.0-<date>.v<commit>` (prior to this change, version uses date as patch: `0.22.<date>-v<commit>`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,10 +71,15 @@ check-sh:
 
 .PHONY: version-vars
 version-vars:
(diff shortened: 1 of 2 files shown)


frhuelsz:
> > Do ADO feeds accept this format?
> 
> needed a fix to make valid semver. it now creates valid semver. does ADO accept it? hard to know without trying???

I generally just do a manual push to the dev feed to validate, because its compliance to the semver spec is very arbitrary

fintelia (approved):

frhuelsz (approved):

PR #596, opened 2026-04-10
engineering: Cross-validation and doc updates for flexible ESP path

frhuelsz:
# 🔍 Description
 
- Assert that the HC and COSI file agree on the ESP mount point.
- Also add a doc on ESP mount overriding.
- Use constants in docbuilder instead of hard-coded values.

Depends on:
- #595


crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -310,6 +310,17 @@ fn validate_esp(os_image: &OsImage, ctx: &EngineContext) -> Result<(), TridentEr
         return Ok(());
     };
 
+    // Ensure the ESP filesystem mount point from the image matches the ESP
+    // mount point in the Host Configuration.
+    if esp_img.mount_point != ctx.esp_mount_point {
+        return Err(TridentError::new(
+            InvalidInputError::EspMountPointMismatch {
+                os_image_esp_mount_point: esp_img.mount_point.display().to_string(),
+                host_config_esp_mount_point: ctx.esp_mount_point.display().to_string(),
```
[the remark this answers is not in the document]
frhuelsz: fixed

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -310,6 +310,17 @@ fn validate_esp(os_image: &OsImage, ctx: &EngineContext) -> Result<(), TridentEr
...
+            },
+        ));
+    }
```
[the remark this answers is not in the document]
frhuelsz: Added `test_validate_esp_mount_point_mismatch` covering: default match (success), non-default match (success), HC→image mismatch, and image→HC mismatch — all asserting `EspMountPointMismatch` with the expected paths. (54c7d4e)

docs/Reference/Host-Configuration/API-Reference/OverrideEspMount.md
bfjelds, line 29: maybe add a note that if this is used, another filesystem if mounted at /boot/efi will not be considered ESP.

docs/Explanation/ESP-Detection.md
```diff
@@ -0,0 +1,218 @@
...
+authoring a Host Configuration for a given distribution, the ESP mount point
+must match the path the distribution's COSI image expects.
+
+## ESP Detection in Host Configuration YAML
+
+### Default Behavior
+
+When a filesystem entry is declared in the Host Configuration and no
+`overrideEspMount` field is specified, Trident applies the `use-default`
+behavior: the filesystem is treated as the ESP **if and only if** its mount
```
bfjelds, line 38: nit nit nit: `it's`

crates/trident_api/schemas/host-config-schema.json
```diff
@@ -1293,7 +1293,7 @@
           }
         },
         "interpreter": {
-          "description": "Binary to run the script with. The default is `/bin/sh`.",
+          "description": "Binary to run the script with. The default is `{{ DEFAULT_SCRIPT_INTERPRETER }}`.",
```
bfjelds: can the variables be substituted here?

frhuelsz: Not with the current version of schemars (0.8), possibly _maybe_ with schemars 1.2

docs/Reference/Host-Configuration/API-Reference/OverrideEspMount.md
frhuelsz: It will be, because of how parsing is scoped. Every filesystem is evaluated independently. We'd need to do something like `initialize()` to be in a context where we can look at all filesystems to derive which one is the real ESP if we have multiple candidates.
If you have a fs with override, and also a fs on `/boot/efi` you'll get an error about having multiple ESPs (with descriptions of each), you'd need to apply `override: block` to the fs in `/boot/efi` to prevent that from being detected as ESP.

In _most_ cases, having `/boot/efi` and _not_ use that as the ESP will most likely be an error. I think adding an extra confirmation of "i am indeed asking you to do this very unusual thing" is reasonable.

docs/Explanation/ESP-Detection.md
frhuelsz: [Copilot] "its" (possessive, no apostrophe) is correct here — it refers to the filesystem's mount point path. "it's" would expand to "it is," which doesn't fit the sentence.

crates/trident_api/schemas/host-config-schema.json
frhuelsz: (At least without doing something more radical like passing the entire file through the template engine, but that sounds a bit too extreme)

docs/Reference/Host-Configuration/API-Reference/OverrideEspMount.md
bfjelds: > If you have a fs with override, and also a fs on /boot/efi you'll get an error about having multiple ESPs

this is the note i was talking about adding

docs/Explanation/ESP-Detection.md
bfjelds: you are totally right ... i was thinking `it is` for some reason.

docs/Reference/Host-Configuration/API-Reference/OverrideEspMount.md
frhuelsz: Ah i see, i thought you meant automatically. Will add.

crates/trident_api/schemas/host-config-schema.json
bfjelds: is it preferrable to have this in the schema to hardcoding in comments elsewhere?

bfjelds: who consumes the schema file?

frhuelsz: The schema itself is unaffected; descriptions are merely for humans. I would expect any human to read the docs instead of this file. In any case, i think it may actually be somewhat easy to solve that by re-arranging where this happens in docbuilder, but i'd probably keep that as a separate PR.

crates/trident_api/src/config/host/os/mod.rs
```diff
@@ -161,14 +161,13 @@ impl FromStr for SelinuxMode {
...
 /// UEFI fallback mode
 ///
-/// UEFI provides a mechanism for booting from an EFI file without
-/// a corresponding boot variable existing in NVRAM. This is known
-/// as the UEFI fallback mode, and it uses a specific file path
-/// (/EFI/BOOT) to locate the fallback bootloader.
+/// UEFI provides a mechanism for booting from an EFI file without a
+/// corresponding boot variable existing in NVRAM. This is known as the UEFI
+/// fallback mode, and it uses a specific file path (/{{
+/// EFI_DEFAULT_BIN_RELATIVE_PATH }}) to locate the fallback bootloader.
```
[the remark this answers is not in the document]
frhuelsz: fixed

crates/trident_api/src/error.rs
```diff
@@ -170,6 +171,18 @@ pub enum InvalidInputError {
     #[error("The provided OS image file is corrupt: {0}")]
     CorruptOsImage(String),
 
+    #[error(
+        "Failed to validate ESP mount point: OS image has ESP mounted at '{os_image_esp_mount_point}' \
+        but Host Configuration has ESP mounted at '{host_config_esp_mount_point}'. If the ESP in the image is \
+        indeed mounted at a location that is NOT the default ESP mount point ('{DEFAULT_ESP_MOUNT_POINT_PATH}'), \
+        mark the correct filesystem as the ESP using the ESP override option in the Host Configuration \
+        filesystem definition."
```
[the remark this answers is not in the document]
frhuelsz: No i don't want to risk this becoming stale if the name changes.

crates/trident_api/schemas/host-config-schema.json
frhuelsz: Handled in:
- #598

bfjelds (approved):

PR #610, opened 2026-04-17
infra: virtdeploy: Extension for ACL images

frhuelsz:
# 🔍 Description
 
Add support for amd64 ACL images with new:
- `-s` auto start the vm after it is created
- `--ignition-config=/path/to/config.ign`

Example:
```bash
./bin/virtdeploy create-one -n acl --no-secure-boot -s --ignition-config=/tmp/acl-ignition.ign --os-disk=/path/to/acl_vm.img
```

### DIFF
--- a/tools/cmd/virtdeploy/main.go
+++ b/tools/cmd/virtdeploy/main.go
@@ -57,18 +57,20 @@ func (c *CleanCmd) Run() error {
 }
 
 type CreateOneCmd struct {
-	Namespace    string `group:"Resource" short:"n" long:"namespace" help:"Namespace to create resources in" default:"${DEFAULT_NAMESPACE}"`
-	Network      string `group:"Resource" short:"N" long:"network" help:"Network to create resources in" default:"${DEFAULT_NETWORK}"`
-	CPUs         uint   `group:"VM" short:"c" long:"cpus" help:"Number of CPUs for the VM" default:"4"`
-	Mem          uint   `group:"VM" short:"m" long:"mem" help:"Memory in GB for the VM" default:"6"`
-	Disks        []uint `group:"VM" short:"d" long:"disk" help:"Disk sizes in GB for the VM" default:"32"`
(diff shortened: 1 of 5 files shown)


tools/pkg/virtdeploy/vm.go
```diff
@@ -183,6 +195,22 @@ func (vm *VirtDeployVM) createDomain(
 				},
 			}},
 			Inputs: []libvirtxml.DomainInput{{Type: "tablet", Bus: "usb"}},
+			Graphics: []libvirtxml.DomainGraphic{{
```
bfjelds: does the new graphics/sounds/videos sections work for arm and amd?

bfjelds: (my recollection is that when i last checked, some of them did not)

frhuelsz: those were there in an old version of the xml, just re-adding them because other parts still refer to spice graphics so libvirt rejects the config, but i have not tested in arm, will limit to just amd for now.

frhuelsz: you're right, this doesn't work on arm, updated to make it conditional to amd

tools/cmd/virtdeploy/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
 		kong.Vars{
 			"DEFAULT_NAMESPACE": DEFAULT_NAMESPACE,
 			"DEFAULT_NETWORK":   DEFAULT_NETWORK,
+			"GOARCH":            runtime.GOARCH,
```
bfjelds, line 38: hopefully this can be overridden.  if our pipelines are an indication, we will want to run emulated arm VMs on amd64.

bfjelds: nevermind...see it as arg below

bfjelds (dismissed):

bfjelds (approved):

PR #617, opened 2026-04-20
Ensure OsRelease alignment between host and cosi for update

bfjelds:
# 🔍 Description
 
Ensure that OsRelease (id and variant_id) match for an update between host and COSI.

### DIFF
--- a/crates/osutils/src/osrelease.rs
+++ b/crates/osutils/src/osrelease.rs
@@ -231,6 +231,12 @@ impl<'de> Deserialize<'de> for OsRelease {
     }
 }
 
+impl Default for &OsRelease {
+    fn default() -> Self {
+        &OsRelease::EMPTY
+    }
+}
+
 /// Represents the contents of the extension-release file of a sysext or
 /// confext.
 ///
--- a/crates/osutils/src/osrelease.rs
+++ b/crates/osutils/src/osrelease.rs
@@ -158,9 +158,24 @@ impl OsRelease {
         }
     }
 
-    // Ensure that the host OsRelease aligns to a COSI file.
-    pub fn ensure_host_alignment(cosi: &OsRelease) -> Result<(), Error> {
-        OsRelease::read()?.ensure_osrelease_alignment(cosi)
+    /// Ensure that the host OsRelease aligns to a COSI file.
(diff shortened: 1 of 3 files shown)


crates/osutils/src/osrelease.rs
```diff
@@ -220,6 +225,30 @@ impl OsRelease {
 
         os_release
     }
+
+    // Ensure that the host OsRelease aligns to a COSI file.
+    fn ensure_osrelease_alignment(&self, other: &OsRelease) -> Result<(), Error> {
```
fintelia: I think this would be clearer if it didn't take `&self` and instead took two well-named arguments

frhuelsz: +1, I'd also suggest renaming to something less ambiguous like ensure same OS type, or same distro

crates/osutils/src/osrelease.rs
```diff
@@ -220,6 +225,30 @@ impl OsRelease {
...
+        // Ensure that the os_releases ids align
+        let id = self.id.clone().unwrap_or("None".to_string());
```
frhuelsz: i think we can compare the options directly, no need to unwrap to "None".
```rust
ensure!(
    self.id == other.id,
    "...",
    self.id.as_deref().unwrap_or("(blank)"),
    other.id.as_deref().unwrap_or("(blank)"),
)
```
 should work too afaik

crates/osutils/src/osrelease.rs
bfjelds: lol, i find it clearer this way (especially in the error message) than what you are suggesting, but happy to adjust.

crates/osutils/src/osrelease.rs
```diff
@@ -158,6 +158,26 @@ impl OsRelease {
         }
     }
 
+    /// Ensure that the host OsRelease aligns to a COSI file.
+    pub fn ensure_matching_distro(host: &OsRelease, cosi: &OsRelease) -> Result<(), Error> {
+        // Ensure that the os_releases ids align
+        ensure!(
+            host.id == cosi.id,
+            "Mismatched OS release: host = {:?}, cosi = {:?}",
```
frhuelsz: nit: maybe clarify that this is the ID field like the message below is stating that it's the VARIANT_ID?

frhuelsz (approved):

fintelia (approved):

PR #618, opened 2026-04-20
Only require grub.cfg for non-uki

bfjelds:
# 🔍 Description
 
Only require grub.cfg for non-uki images

### DIFF
--- a/crates/trident/src/subsystems/esp.rs
+++ b/crates/trident/src/subsystems/esp.rs
@@ -254,8 +254,8 @@ fn copy_file_artifacts(
 
     // Generate list of filepaths to the boot files. Pass in the temp dir path where the image is
     // mounted to as an argument
-    let boot_files =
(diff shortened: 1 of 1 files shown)


crates/trident/src/subsystems/esp.rs
```diff
@@ -621,17 +624,22 @@ fn generate_boot_filepaths(temp_mount_dir: &Path) -> Result<Vec<PathBuf>, Error>
...
-    );
-    paths.push(selected_grub_config_path);
+
+    if let Some(selected_grub_config_path) = selected_grub_config_path {
+        debug!(
+            "Using GRUB configuration file '{GRUB2_CONFIG_FILENAME}' from '{}'",
+            selected_grub_config_path.display()
+        );
+        paths.push(selected_grub_config_path);
+    }
```
fintelia: Could we instead wrap this entire block in `if !is_uki { ... }`?

fintelia (approved):

PR #619, opened 2026-04-21
In A/B update validation, check /usr mount points if root isn't A/B

fintelia:


crates/trident_api/src/config/host/storage/storage_graph/graph.rs
```diff
@@ -228,6 +228,18 @@ impl StorageGraph {
         let (fs_idx, _) = self.filesystem_node_by_mount_point(mount_path)?;
         self.backing_verity_device(fs_idx).map(|(_, dev)| dev)
     }
+
+    pub fn root_fs_is_ab(&self) -> bool {
```
frhuelsz: nit, may need UT

crates/trident/src/engine/rollback.rs
```diff
@@ -574,97 +574,122 @@ fn compare_root_device_paths(
...
+            .get_root_block_device_id()
+            .ok_or_else(|| TridentError::new(ServicingError::GetRootBlockDeviceId))?;
+
+        validate_ab_active_volume_internal(
+            ctx,
+            root_device_path,
+            root_device_id,
+            ctx.storage_graph.root_fs_is_verity(),
+        )
+    } else {
```
frhuelsz: i think we should gate on this being ACL, generally we probably don't want to allow this otherwise

crates/trident/src/engine/rollback.rs
```diff
@@ -574,97 +574,122 @@ fn compare_root_device_paths(
 /// volume set in the Host Status/context accurately reflects the actual active volume, i.e., the
 /// device that the firmware booted from.
 pub(crate) fn validate_ab_active_volume(ctx: &EngineContext) -> Result<(), TridentError> {
-    let root_device_path = block_devices::get_root_device_path()?;
-    validate_ab_active_volume_internal(ctx, root_device_path)
+    if ctx.storage_graph.root_fs_is_ab() {
```
frhuelsz: might be worth debug logging which arm of this if statement we followed

frhuelsz (approved):

PR #622, opened 2026-04-24
infra: virt-firmware 26.4 breaks our tests

bfjelds:
# 🔍 Description
 
virt-firmware recently released 26.4 which breaks our tests' invocation of `virt-fw-vars --enroll-cert`

### DIFF
--- a/.pipelines/templates/stages/testing_vm/netlaunch-prep.yml
+++ b/.pipelines/templates/stages/testing_vm/netlaunch-prep.yml
@@ -58,7 +58,7 @@ steps:
 
       # Use `-E` to preserve PipAuthenticate connection to
       # internal feed replacing PyPI
-      sudo -E pip3 install virt-firmware -vvv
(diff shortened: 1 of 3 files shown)


frhuelsz (dismissed):

frhuelsz:
I wonder if we should starts writing these down to a `requirements.txt` file in `tools/` or somewhere :thinking:

bfjelds:
> I wonder if we should starts writing these down to a `requirements.txt` file in `tools/` or somewhere 🤔

updated docs/Development/Building/Dependencies.md.

frhuelsz:
> > I wonder if we should starts writing these down to a `requirements.txt` file in `tools/` or somewhere 🤔
> 
> updated docs/Development/Building/Dependencies.md.

I mean this https://pip.pypa.io/en/stable/reference/requirements-file-format/, so that we have a single location with the list of things and their versions. (but happy to leave that for later)

frhuelsz (approved):

PR #634, opened 2026-05-08
storm: improve boot timeout diagnostics and increase timeout

bfjelds:
## Problem

Scale test run 1112556 had 3 VM timeout failures during update-loop. Root cause analysis was impossible because serial log content is not printed in pipeline output and logs are only in multi-GB artifacts.

### Serial log tail output
analyzeSerialLog() now prints the last 100 lines of the serial log on boot failure.

### Boot failure pattern detection
- Initramfs emergency shell
- Kernel panic
- GRUB boot error

### Timeout increase (120s to 180s)
Extra margin for CPU contention at scale.

## Related
- Addresses diagnostic gap from scale test run 1112556
- Complements PR #632 (multi-IP DHCP fix)


bfjelds:
i think this is an AI swing and miss.

PR #669, opened 2026-06-01
feat(uki): activate correct verity addon for target A/B usr partition (ACL only)

bfjelds:
ACL images ship with PARTUUID-based verity addons — for both A and B partitions stored in acl/uki-addons/ on the ESP, with partition A active by default. During an A/B update, trident must swap the active addon to match the target partition so the new UKI boots with the correct verity partition identity.

Add activate_verity_addon_for_target_volume() which:
- Checks for ACL verity addon templates on the image ESP
- Copies the correct partition addon into the staged addon directory
- Is a silent no-op for non-ACL images (no template dir)
- Errors if template dir exists but the selected partition is missing

Called from copy_file_artifacts() after stage_uki_on_esp(), gated on ctx.image_distro().is_acl() to ensure only ACL images are affected.

### DIFF
--- a/crates/trident/src/engine/boot/uki.rs
+++ b/crates/trident/src/engine/boot/uki.rs
@@ -937,9 +937,7 @@ mod tests {
     // ── activate_verity_addon_for_target_volume tests ──────────────────────
 
     /// Helper: creates a mock image ESP with verity addon templates.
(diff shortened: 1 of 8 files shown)


bfjelds:
replaced with https://github.com/microsoft/trident/pull/673

PR #673, opened 2026-06-04
engineering: ACL COSI: handle shared BTRFS UUIDs and ESP space management

bfjelds:
## Summary

Enables trident to successfully perform A/B updates on ACL when the COSI image shares BTRFS filesystem UUIDs with the currently active OS. Also adds pre-staging ESP cleanup to prevent "no space left on device" failures during repeated updates.

## Related PRs

* acl-scripts: provide A/B command line UKIs: https://dev.azure.com/mariner-org/ACL/_git/acl-scripts/pullrequest/27615
* acl-pipelines: create single VHD/COSI combo, run 5 updates: https://dev.azure.com/mariner-org/ACL/_git/acl-pipelines/pullrequest/27673

## Combined Validation

https://dev.azure.com/mariner-org/ACL/_build/results?buildId=1132645

### ACL BTRFS UUID handling
- **Skip duplicate FS UUID check for ACL** — ACL uses PARTUUID-based boot (not GRUB `search --fs-uuid`), so shared UUIDs are safe
- **Validate ACL duplicate UUIDs via verity root hash** — ensures the duplicate is genuinely the same image, not a different OS masquerading with the same UUID
- **Runtime BTRFS UUID bind-mount fallback** — when kernel BTRFS rejects a mount due to duplicate UUIDs, bind-mount from the already-mounted filesystem instead
- **Verify verity root hash before bind-mount** — safety check to confirm the active and target partitions have matching verity hashes

### UKI enhancements
- **Activate correct verity addon for target A/B slot** — selects the right `verity.addon.efi` based on which slot is being updated
- **Include UKI addons in `findUkiEntries` COSI metadata** — ensures addon files are discovered during COSI parsing
- **Search UKI addon cmdlines for `usrhash=` parameter** — extracts verity root hash from addon kernel cmdline

### ESP space management
- **Pre-staging UKI cleanup** — removes old target-slot UKIs before staging new ones, preventing ESP overflow on 128 MB partitions
- **Multi-OS scoping** — cleanup matches exact slot+os-index (e.g. `azla0`) not just slot letter, safe for multiboot
- **Original UKI cleanup guarded by `install_index == 0`** — only the OS that placed the original UKI can remove it

### Internal testing support
- **`forceAbUpdate` internal param** — bypasses SHA384 identity check, allowing the same COSI to be applied repeatedly for A↔B cycle testing

## Testing

- Build verified clean on Linux (`cargo check -p trident`, `cargo fmt`)
- acl-pipelines branch `user/bfjelds/single-acl-build` updated with 5-cycle A↔B test and ESP diagnostics
- Pipeline validation in progress


crates/trident/src/engine/newroot.rs
```diff
@@ -202,6 +216,38 @@ impl NewrootMount {
...
+                // ACL-specific: if the staging device has a BTRFS filesystem UUID that
+                // collides with the active USR partition, bind-mount from the host's
+                // /usr instead. The verity-protected filesystem is read-only and the
+                // content is identical when UUIDs match, so the bind mount provides
+                // equivalent content for chroot provisioning.
+                if let Some(ref collision_uuid) = acl_collision_uuid {
+                    if fs_type == Some(RealFilesystemType::Btrfs)
+                        && block_device.fsuuid.as_ref() == Some(collision_uuid)
+                    {
+                        let active_usr = Path::new("/usr");
```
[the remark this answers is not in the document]
bfjelds: Good catch. Fixed in f27b359b — the UUID collision workaround now also checks that the mount point is `/usr` before applying the bind-mount. Other BTRFS partitions that happen to share the UUID will be mounted normally.

tools/cmd/mkcosi/generator/generator.go
```diff
@@ -1085,17 +1085,63 @@ func findUkiEntries(espMountPath string, espMountPoint string) []metadata.System
...
+// these addons at boot and appends their cmdline args to the UKI's own cmdline.
+func findUkiAddons(ukiDir string, ukiName string, espMountPoint string) []metadata.UkiAddon {
+	addonDirName := ukiName + ".extra.d"
+	addonHostDir := filepath.Join(ukiDir, addonDirName)
+
+	addonEntries, err := os.ReadDir(addonHostDir)
+	if err != nil {
+		// No .extra.d directory is normal for UKIs without addons
+		return nil
+	}
```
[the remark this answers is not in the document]
bfjelds: Good catch. Fixed in f27b359b — now only `os.ErrNotExist` is silently ignored. Other `ReadDir` errors are logged as warnings so permission issues or corrupt mounts surface in diagnostics.

crates/trident/src/subsystems/esp.rs
```diff
@@ -288,8 +288,31 @@ fn copy_file_artifacts(
         // Prepare ESP directory structure for UKI boot
         uki::prepare_esp_for_uki(mount_point, &ctx.esp_mount_path)?;
 
+        // Clean up old UKIs for the target slot before staging the new one.
+        // Pre-staging cleanup is necessary because the ESP may not have space
+        // for both old and new UKIs simultaneously (128 MB constraint). The
+        // trade-off: a crash between cleanup and staging removes the target
+        // slot's old UKI with no replacement, but the active slot's UKI is
+        // preserved so the system remains bootable via A/B fallback.
+        uki::cleanup_ukis_before_staging(ctx, mount_point, &ctx.esp_mount_path)?;
```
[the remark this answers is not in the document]
bfjelds: This is safe as-is. On clean install the ESP UKI directory does not exist yet, so `cleanup_ukis_before_staging` returns `Ok(())` immediately at line 231-233 (the `!esp_uki_directory.exists()` guard). No UKIs are deleted. The function only removes UKIs that are already present, making it a no-op for clean installs.

crates/trident/src/engine/boot/uki.rs
```diff
@@ -160,6 +182,98 @@ pub fn prepare_esp_for_uki(root_mount_point: &Path, esp_mount_path: &Path) -> Re
...
+    //    - Trident already manages the active slot (proving the original is
+    //      superseded as rollback).
+    //    In multiboot, install_index > 0 never owns the original UKI — it has
+    //    OS 0's partition refs baked in and can't boot other OS instances.
+    let has_active_slot_uki = trident_ukis
+        .iter()
+        .any(|(_index, suffix, _)| *suffix == active_suffix);
+
+    if has_active_slot_uki && ctx.install_index == 0 {
+        let non_trident_ukis = enumerate_non_trident_managed_ukis(&esp_uki_directory)?;
```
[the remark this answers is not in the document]
bfjelds: Good catch. Fixed in 8deafef0 — non-Trident-managed UKIs are now only deleted when Trident has established UKIs for both the active and target slots (`has_active_slot_uki && has_target_slot_uki`). This preserves the original install UKI as a rollback target during the critical transition period when only one slot is Trident-managed.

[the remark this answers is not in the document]
bfjelds: Correction: the previous fix (8deafef0) was reverted in 2006b62c. The original logic is correct.

The ESP is ~125 MB and each UKI is ~50 MB. Requiring both slots to have Trident UKIs before deleting the original means 3 UKIs on disk simultaneously (active + original + new target = 150 MB), which exceeds ESP capacity and breaks staging.

The original condition (\\has_active_slot_uki\\ only) is the intended design: staging already breaks rollback to the original UKI, so deleting it pre-staging to free space is the accepted trade-off. Added a comment explaining this rationale.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -259,6 +275,121 @@ fn validate_filesystem_uniqueness(
...
+/// ACL BTRFS images may intentionally share filesystem UUIDs between the A and B
+/// slots when built from the same source. This is safe only when:
+///
+/// 1. The duplicate is on the `/usr` mount point (ACL's verity-protected USR partition)
+/// 2. The staging image has a verity root hash in its COSI metadata
+/// 3. The active system's `/proc/cmdline` has a matching `usrhash=` parameter
+/// 4. The normalized root hashes match — proving byte-identical content via merkle tree
+///
+/// If COSI partition metadata is available, the function also validates that the
+/// staging USR partition's PARTUUID matches a known ACL USR slot.
```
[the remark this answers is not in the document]
bfjelds: Good catch. Fixed in 859294ed — the doc comment now clarifies the PARTUUID check scans all partitions for a known ACL USR PARTUUID (which is equivalent to checking the USR partition specifically, since GPT PARTUUIDs are globally unique).

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -189,6 +191,9 @@ fn validate_filesystems(os_image: &OsImage, ctx: &EngineContext) -> Result<(), T
 
 /// Validates that all filesystems within an OS image have unique FS UUIDs. Additionally, validates
 /// that A/B volume pairs have distinct FS UUIDs.
+///
+/// The A/B cross-check is skipped for ACL, which uses identical FS UUIDs across A/B slots by
+/// design (partitions are distinguished by PARTUUID instead).
```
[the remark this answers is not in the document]
bfjelds: Good catch. Fixed in 859294ed — the doc comment now accurately describes the behavior: ACL duplicates are conditionally allowed after verity verification, not skipped entirely.

tools/cmd/mkcosi/metadata/metadata.go
```diff
@@ -65,6 +65,12 @@ type SystemDBootEntry struct {
 	Path    string               `json:"path"`
 	Cmdline string               `json:"cmdline"`
 	Kernel  string               `json:"kernel"`
+	Addons  []UkiAddon           `json:"addons,omitempty"`
```
frhuelsz: breaks spec
https://microsoft.github.io/trident/docs/v0.21.0/Reference/Composable-OS-Image#systemdbootentry-object

crates/trident/src/engine/boot/uki.rs
```diff
@@ -25,11 +25,33 @@ pub const TMP_UKI_NAME: &str = "vmlinuz-0.efi.staged";
 pub const UKI_DIRECTORY: &str = formatcp!("{ESP_EFI_DIRECTORY}/Linux");
 const TMP_UKI_ADDON_DIR_NAME: &str = formatcp!("{TMP_UKI_NAME}{UKI_ADDON_DIR_SUFFIX}");
 
+/// Slot identifier embedded in trident-managed UKI filenames.
+const UKI_SLOT_A: &str = "azla";
+const UKI_SLOT_B: &str = "azlb";
```
frhuelsz: worth considering these pre-existing consts in constants.rs, but unsure if casing matters:

```rust
/// Azure Linux Install ID Prefix
pub const AZURE_LINUX_INSTALL_ID_PREFIX: &str = "AZL";

/// A/B Volume A Name
pub const AB_VOLUME_A_NAME: &str = "A";

/// A/B Volume B Name
pub const AB_VOLUME_B_NAME: &str = "B";
```

crates/trident/src/engine/boot/uki.rs
```diff
@@ -160,6 +182,102 @@ pub fn prepare_esp_for_uki(root_mount_point: &Path, esp_mount_path: &Path) -> Re
     Ok(())
 }
 
+/// Removes a UKI file and its associated addon directory (if present).
+/// Treats `NotFound` as success for idempotency — a partially-failed previous
+/// cleanup should not prevent the addon directory from being cleaned.
+fn remove_uki_and_addons(path: &Path) -> Result<(), Error> {
+    match fs::remove_file(path) {
+        Ok(()) => {}
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
```
frhuelsz: nit: import directly

crates/trident/src/engine/boot/uki.rs
```diff
@@ -160,6 +182,102 @@ pub fn prepare_esp_for_uki(root_mount_point: &Path, esp_mount_path: &Path) -> Re
...
+            trace!("UKI file '{}' already removed, continuing", path.display());
+        }
+        Err(e) => {
+            return Err(anyhow::Error::new(e)
+                .context(format!("Failed to remove UKI file '{}'", path.display())));
+        }
+    }
```
frhuelsz, line 209: nit:
```suggestion
    match fs::remove_file(path) {
        Err(e) if e.kind() == ErrorKind::NotFound => {
            trace!("UKI file '{}' already removed, continuing", path.display());
        }
        e => e.with_context(|| format!("Failed to remove UKI file '{}'", path.display()))?,
    }
```

crates/trident/src/subsystems/esp.rs
```diff
@@ -288,8 +288,31 @@ fn copy_file_artifacts(
...
+        if ctx.image_distro().is_acl() {
+            if let Some(target_volume) = ctx.get_ab_update_volume() {
+                uki::activate_verity_addon_for_target_volume(
+                    temp_mount_dir,
+                    mount_point,
+                    &ctx.esp_mount_path,
+                    target_volume,
+                )?;
+            }
+        }
```
frhuelsz, line 315: have we committed to this approach on ACL?

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -29,6 +29,8 @@ use crate::{
     osimage::{OsImage, OsImageFileSystemType},
 };
 
+use sysdefs::osuuid::OsUuid;
```
frhuelsz: nit: move to correct import block

crates/trident/src/engine/acl.rs
```diff
@@ -0,0 +1,107 @@
...
+//!
+//! ACL uses fixed PARTUUIDs for USR A/B partitions and a verity addon that
+//! places the root hash in the kernel command line as `usrhash=<hex>`.
+
+use std::fs;
+
+// ACL UKI disk layout defines fixed PARTUUIDs for the USR A/B data partitions.
+// These are from acl-scripts disk_layout_uki.json.
+pub const ACL_USR_A_PARTUUID: &str = "7130c94a-213a-4e5a-8e26-6cce9662f132";
+pub const ACL_USR_B_PARTUUID: &str = "e03dd35c-7c2d-4a47-b3fe-27f15780a57c";
```
frhuelsz: given their scope i feel these should live in constants.rs

crates/trident/src/engine/acl.rs
```diff
@@ -0,0 +1,107 @@
+//! ACL (Azure Container Linux) UKI-specific constants and helpers.
```
frhuelsz: this file reads to me like something that should live in osutils, something like verity_roothash.rs perhaps?

It may also be more convenient to not handle bare Strings and define a `struct VerityRootHash(String)` that is returned by `read_active_usr_roothash` and implements comparison against itself and `impl AsRef<str>`

crates/trident/src/engine/ab_update.rs
```diff
@@ -59,13 +62,25 @@ pub(super) fn stage_update(
...
 
     storage::initialize_block_devices(&ctx)?;
+
+    // Extract the staging USR verity root hash from the COSI image metadata.
+    // This is used to cryptographically verify that the active and staging USR
+    // partitions have identical content before allowing a bind-mount workaround
+    // for the BTRFS kernel UUID collision on ACL.
+    let staging_usr_roothash = ctx.image.as_ref().and_then(|img| {
+        img.filesystems()
+            .find(|fs| fs.mount_point == Path::new("/usr"))
```
frhuelsz: nit, prefer the pre-defined const for /usr

crates/trident/src/subsystems/esp.rs
bfjelds: have we committed to anything?

crates/trident/src/engine/ab_update.rs
frhuelsz: iirc, I think you can compare directly without having to `Path::new`

crates/trident/src/engine/newroot.rs
```diff
@@ -339,6 +386,103 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
...
+fn detect_acl_btrfs_uuid_collision(
+    update_volume: AbVolumeSelection,
+    staging_usr_roothash: Option<&str>,
+) -> Option<OsUuid> {
+    let (active_partuuid, update_partuuid) = match update_volume {
+        AbVolumeSelection::VolumeA => (ACL_USR_B_PARTUUID, ACL_USR_A_PARTUUID),
+        AbVolumeSelection::VolumeB => (ACL_USR_A_PARTUUID, ACL_USR_B_PARTUUID),
+    };
+
+    let active_path = format!("/dev/disk/by-partuuid/{active_partuuid}");
```
frhuelsz: nit: osutils::block_devices::part_uuid_path

crates/trident/src/engine/newroot.rs
```diff
@@ -339,6 +386,103 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
...
+/// verity device cannot be mounted.
+///
+/// This function checks whether the active and update USR partitions (identified by
+/// their well-known ACL PARTUUIDs) have the same BTRFS filesystem UUID. If so, it
+/// returns the colliding UUID so the caller can substitute a bind mount from the
+/// active `/usr`.
+///
+/// Returns `None` if:
+/// - The system is not ACL (PARTUUIDs not found)
+/// - The partitions don't have BTRFS filesystems
```
frhuelsz: i don't know what happens ix an ext4 partition and a btrfs partition have the same UUID

tools/cmd/mkcosi/metadata/metadata.go
bfjelds: Good catch -- fixed in d28b9d9f. Instead of extending the metadata schema, addon cmdlines are now folded into the entry Cmdline field during COSI generation (mirroring what systemd-stub does at boot). The addons field and UkiAddon type have been removed from both Go and Rust.

crates/trident/src/engine/boot/uki.rs
bfjelds: Fixed in 99f4b1e4 -- imported ErrorKind directly.

crates/trident/src/engine/boot/uki.rs
bfjelds: Applied -- much cleaner, thanks. Fixed in 99f4b1e4.

crates/trident/src/subsystems/storage/osimage.rs
bfjelds: Moved sysdefs import up into the external crate block in 99f4b1e4.

crates/trident/src/engine/acl.rs
bfjelds: Agreed -- moved ACL_USR_A_PARTUUID and ACL_USR_B_PARTUUID to constants.rs and re-exported from acl.rs for backward compatibility. Fixed in 99f4b1e4.

crates/trident/src/engine/newroot.rs
bfjelds: is that an ACL possibility?

crates/trident/src/engine/boot/uki.rs
bfjelds: Good idea to keep these coupled. Added UKI_SLOT_A/UKI_SLOT_B to constants.rs next to the parent consts (AZURE_LINUX_INSTALL_ID_PREFIX, AB_VOLUME_A/B_NAME) with doc comments linking them. Casing does matter (lowercase in filenames), so they are separate consts rather than computed at runtime. Fixed in 22d01294.

crates/trident/src/engine/acl.rs
bfjelds: Done in 22d01294. Created osutils::verity_roothash with a VerityRootHash(String) newtype that handles normalization, case-insensitive comparison via PartialEq, preview formatting, and /proc/cmdline parsing. The engine::acl module now just re-exports the type and ACL PARTUUID constants.

crates/trident/src/engine/newroot.rs
bfjelds: Good call — switched to block_devices::part_uuid_path in 8c581394.

crates/trident/src/engine/ab_update.rs
bfjelds: Fixed in bb13d6ff — now using USR_MOUNT_POINT_PATH from constants.

crates/trident/src/engine/newroot.rs
```diff
@@ -339,6 +388,115 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
     }
 }
 
+use osutils::verity_roothash::VerityRootHash;
+use trident_api::constants::{ACL_USR_A_PARTUUID, ACL_USR_B_PARTUUID};
```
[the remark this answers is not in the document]
bfjelds: Fixed in 21a9724e — moved imports to top of file.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+    #[test]
+    fn test_acl_duplicate_uuid_matching_hash_success() {
+        let mock = acl_mock_image(Some(ACL_TEST_HASH));
+        let os_image = OsImage::mock(mock);
+        let result = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(ACL_TEST_HASH.to_string()),
+        );
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+    #[test]
+    fn test_acl_duplicate_uuid_matching_hash_case_insensitive() {
+        let mock = acl_mock_image(Some(ACL_TEST_HASH));
+        let os_image = OsImage::mock(mock);
+        let result = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(ACL_TEST_HASH.to_uppercase()),
+        );
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+
+    #[test]
+    fn test_acl_duplicate_uuid_matching_hash_success() {
+        let mock = acl_mock_image(Some(ACL_TEST_HASH));
+        let os_image = OsImage::mock(mock);
+        let result = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(ACL_TEST_HASH.to_string()),
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/engine/newroot.rs
```diff
@@ -339,6 +386,115 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
...
+        AbVolumeSelection::VolumeA => (ACL_USR_B_PARTUUID, ACL_USR_A_PARTUUID),
+        AbVolumeSelection::VolumeB => (ACL_USR_A_PARTUUID, ACL_USR_B_PARTUUID),
+    };
+
+    let active_path = block_devices::part_uuid_path(
+        uuid::Uuid::parse_str(active_partuuid).expect("ACL PARTUUID constant is valid"),
+    );
+    let update_path = block_devices::part_uuid_path(
+        uuid::Uuid::parse_str(update_partuuid).expect("ACL PARTUUID constant is valid"),
+    );
```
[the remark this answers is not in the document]
bfjelds: These are compile-time-known constant strings — an expect() is appropriate here. If they ever become invalid, the panic message makes the cause obvious. Returning a Result would add error-handling complexity for something that can never fail in practice.

tools/cmd/mkcosi/generator/generator.go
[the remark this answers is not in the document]
bfjelds: Valid observation, but this is pre-existing behavior outside the scope of this PR.

crates/trident/src/subsystems/esp.rs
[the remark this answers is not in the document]
bfjelds: Pre-existing behavior outside the scope of this PR.

crates/trident/src/engine/boot/uki.rs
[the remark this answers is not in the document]
bfjelds: Pre-existing behavior outside the scope of this PR.

crates/trident/src/engine/boot/uki.rs
```diff
@@ -160,6 +182,102 @@ pub fn prepare_esp_for_uki(root_mount_point: &Path, esp_mount_path: &Path) -> Re
...
+            )
+        })?;
+    }
+    Ok(())
+}
+
+/// Cleans up old UKIs for the target slot before staging a new one, freeing
+/// ESP space.  This runs for all UKI-based A/B updates (not just ACL) because
+/// ESP space constraints are a general concern.
+///
```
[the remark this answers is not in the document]
bfjelds: Pre-existing behavior outside the scope of this PR.

crates/trident/src/engine/boot/uki.rs
```diff
@@ -160,6 +193,98 @@ pub fn prepare_esp_for_uki(root_mount_point: &Path, esp_mount_path: &Path) -> Re
...
+    }
+    let addon_dir = uki::uki_addon_dir(path);
+    if addon_dir.exists() && addon_dir.is_dir() {
+        fs::remove_dir_all(&addon_dir).with_context(|| {
+            format!(
+                "Failed to remove UKI addon directory '{}'",
+                addon_dir.display()
+            )
+        })?;
+    }
```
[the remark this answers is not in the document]
bfjelds: Edge case with a non-directory at the addon path — pre-existing behavior, not introduced by this PR.

tools/cmd/mkcosi/generator/generator.go
```diff
@@ -1085,17 +1086,71 @@ func findUkiEntries(espMountPath string, espMountPoint string) []metadata.System
...
+		addonHostPath := filepath.Join(addonHostDir, aName)
+		addonFsPath := filepath.Join(espMountPoint, "EFI", "Linux", addonDirName, aName)
+
+		cmdline := extractUkiSection(addonHostPath, ".cmdline")
+		if cmdline == "" {
+			log.WithField("path", addonFsPath).Debug("Skipping addon with no .cmdline section")
+			continue
+		}
+
+		log.WithField("path", addonFsPath).Debugf("Found UKI addon cmdline: %s", cmdline)
```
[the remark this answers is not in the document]
bfjelds: The base UKI cmdline content is already logged elsewhere in the codebase at similar debug level. The addon cmdlines contain the same type of content (verity hashes, mount options). This is consistent with existing logging.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: The doc comment is accurate as written — it says duplicates are 'permitted if verity verification confirms byte-identical content', which is exactly what the code does.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: The doc comment (lines 287-290) already says 'whether any partition's PARTUUID matches a known ACL USR slot', which accurately describes the implementation.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+    fn test_acl_duplicate_uuid_wrong_mount_point() {
+        let mut mock = acl_mock_image(Some(ACL_TEST_HASH));
+        mock.images[0].mount_point = PathBuf::from("/");
+        let os_image = OsImage::mock(mock);
+        let err = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(ACL_TEST_HASH.to_string()),
+        )
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+    #[test]
+    fn test_acl_duplicate_uuid_no_staging_verity_hash() {
+        let mock = acl_mock_image(None);
+        let os_image = OsImage::mock(mock);
+        let err = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(ACL_TEST_HASH.to_string()),
+        )
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+    fn test_acl_duplicate_uuid_mismatched_hash() {
+        let mock = acl_mock_image(Some(ACL_TEST_HASH));
+        let os_image = OsImage::mock(mock);
+        let different_hash = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
+        let err = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(different_hash.to_string()),
+        )
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+    #[test]
+    fn test_acl_duplicate_uuid_empty_active_hash() {
+        let mock = acl_mock_image(Some(ACL_TEST_HASH));
+        let os_image = OsImage::mock(mock);
+        let err = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(String::new()),
+        )
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+            matches!(
+                err.kind(),
+                ErrorKind::InvalidInput(
+                    InvalidInputError::DuplicateFsUuidAclVerificationFailed { reason, .. }
+                ) if reason.contains("usrhash=")
+            ),
+            "expected empty active hash error, got: {:?}",
+            err.kind()
+        );
+    }
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+
+    #[test]
+    fn test_acl_duplicate_uuid_matching_hash_case_insensitive() {
+        let mock = acl_mock_image(Some(ACL_TEST_HASH));
+        let os_image = OsImage::mock(mock);
+        let result = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(ACL_TEST_HASH.to_uppercase()),
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+    #[test]
+    fn test_acl_duplicate_uuid_wrong_mount_point() {
+        let mut mock = acl_mock_image(Some(ACL_TEST_HASH));
+        mock.images[0].mount_point = PathBuf::from("/");
+        let os_image = OsImage::mock(mock);
+        let err = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(ACL_TEST_HASH.to_string()),
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+
+    #[test]
+    fn test_acl_duplicate_uuid_no_staging_verity_hash() {
+        let mock = acl_mock_image(None);
+        let os_image = OsImage::mock(mock);
+        let err = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(ACL_TEST_HASH.to_string()),
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+    #[test]
+    fn test_acl_duplicate_uuid_mismatched_hash() {
+        let mock = acl_mock_image(Some(ACL_TEST_HASH));
+        let os_image = OsImage::mock(mock);
+        let different_hash = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
+        let err = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(different_hash.to_string()),
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -1301,6 +1433,185 @@ mod tests {
...
+
+    #[test]
+    fn test_acl_duplicate_uuid_empty_active_hash() {
+        let mock = acl_mock_image(Some(ACL_TEST_HASH));
+        let os_image = OsImage::mock(mock);
+        let err = validate_acl_duplicate_uuid(
+            &os_image,
+            Path::new("/usr"),
+            &OsUuid::Uuid(Uuid::parse_str("a1a2a3a4b1b2c1c2d1d2d3d4d5d6d7d8").unwrap()),
+            Some(String::new()),
```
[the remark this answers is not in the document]
bfjelds: Fixed in f87bf946 — all test callsites now use VerityRootHash::new() instead of raw strings.

crates/trident/src/engine/newroot.rs
frhuelsz: _probaly no_, a quick search says i't probably not an issue because of different drivers, perhaps not a real concern here

crates/trident/src/engine/boot/uki.rs
```diff
@@ -16,7 +17,12 @@ use osutils::{
 use trident_api::error::{
     InternalError, ReportError, ServicingError, TridentError, TridentResultExt,
 };
-use trident_api::{constants::ESP_EFI_DIRECTORY, status::AbVolumeSelection};
+use trident_api::{
```
bfjelds, line 17: consolidate with other `use trident_api`

crates/osutils/src/verity_roothash.rs
```diff
@@ -0,0 +1,125 @@
...
+//! hashes, including case-insensitive comparison, preview formatting, and
+//! extraction from `/proc/cmdline`.
+
+use std::fmt;
+use std::fs;
+
+/// A dm-verity root hash, normalized to lowercase with whitespace trimmed.
+///
+/// An empty string is not a valid hash — comparisons against or with an empty
+/// `VerityRootHash` always return `false`.
```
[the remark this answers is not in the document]
bfjelds: Valid — fixed in 47e2c65e. Updated the doc comment to accurately describe the type invariant: `new()` returns `None` for empty input, so a valid `VerityRootHash` always contains a non-empty string.

crates/trident/src/subsystems/esp.rs
frhuelsz: Should we take this out and revisit in 0.25?

tools/cmd/mkcosi/generator/generator.go
```diff
@@ -1085,17 +1086,71 @@ func findUkiEntries(espMountPath string, espMountPoint string) []metadata.System
...
 
+// findUkiAddonCmdlines scans the <ukiName>.extra.d/ directory for UKI addon
+// files (*.addon.efi) and extracts their .cmdline PE sections. Returns the
+// cmdline strings so the caller can fold them into the UKI entry's Cmdline,
+// mirroring what systemd-stub does at boot time.
+func findUkiAddonCmdlines(ukiDir string, ukiName string, espMountPoint string) []string {
+	addonDirName := ukiName + ".extra.d"
+	addonHostDir := filepath.Join(ukiDir, addonDirName)
+
+	addonEntries, err := os.ReadDir(addonHostDir)
```
fintelia, line 1120: I think we need to sort these by filename to make sure the order matches what systemd-boot will do (but please confirm exactly how the sorting works)

bfjelds: sorting the contents of the extra.d folder?  makes sense if/when we have multiple addons in there.  should only be an issue if encryption exists thought, right?  is encryption an ACL scenario?

crates/trident/src/engine/newroot.rs
```diff
@@ -339,6 +390,112 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
...
+
+    let active_dev = lsblk::get(&active_path).ok()?;
+    let update_dev = lsblk::get(&update_path).ok()?;
+
+    if active_dev.fstype.as_deref()? != "btrfs" {
```
frhuelsz: nit:
this this is more of a limitation of the original lsblk::get impl, but in lieu of exploding the scope of the fix, prefer either convert the type to KernelFilesystemType:
```rust
KernelFilesystemType::from(active_dev.fstype.as_deref()?) != RealFilesystemType::Btrfs.as_kernel()
# preferred, i think?
```
or do str comparison of the enum variant:
```rust
active_dev.fstype.as_deref()? != RealFilesystemType::Btrfs.as_kernel().name()

File: tools/cmd/mkcosi/generator/generator.go (hunk 38 above)
Comment (bfjelds) [reply tools/cmd/mkcosi/generator/generator.go:1120]: oh ... the ordering of options in the "composite command line" will vary based on sort order too.  maybe we should mandate a naming convention with a numerical prefix for addon files?

File: crates/trident/src/engine/newroot.rs [hunk 40]
```diff
@@ -339,6 +390,112 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
...
+    // guarantee that the filesystems are byte-identical, not just a UUID match.
+    let Some(staging_hash) = staging_usr_roothash else {
+        // No staging hash available — cannot verify content identity.
+        // Refusing the bind-mount is the safe choice: proceeding without
+        // verification could mount different content at /usr.
+        warn!(
+            "No staging USR verity root hash provided. \
+             Refusing bind-mount — cannot verify content identity."
+        );
+        return None;
```
frhuelsz: at this point we know we are going to fail no? should we return a more controlled error instead of letting the error happen when we attempt the mount?

tools/cmd/mkcosi/generator/generator.go
bfjelds: AI says systemd processes extra.d: "Sort: Case-sensitive strcmp16 within each directory — raw UTF-16 code unit comparison (effectively C locale ASCII: 0-9 < A-Z < a-z)"

so we could do the same or impose a naming convention on ACL?

crates/trident/src/engine/newroot.rs
```diff
@@ -339,6 +390,112 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
...
+    };
+
+    let staging = match VerityRootHash::new(staging_hash) {
```
frhuelsz: nit:
```suggestion
    let Some(staging) = VerityRootHash::new(staging_hash) else {
        warn!("Staging USR verity root hash is empty. Refusing bind-mount.");
        return None;
    };
```

tools/cmd/mkcosi/generator/generator.go
bfjelds: sounds like AI thinks we need to do some fixing in prediction code:

```
In pcrlock.rs around line 753, collect entries into a Vec, sort by filename, then iterate:

 let mut addon_entries: Vec<_> = fs::read_dir(&uki_addon_path)
     .context("Failed to read UKI addons directory")?
     .filter_map(|e| e.ok())
     .filter(|e| uki::is_uki_addon_file(&e.path()))
     .collect();
 addon_entries.sort_by_key(|e| e.file_name());

Rust's OsString sort is byte-level ascending, which matches strcmp16 for ASCII filenames (and systemd rejects non-ASCII addon filenames anyway).
```

AI thinks the go code is ok:

```
 Go's os.ReadDir() returns entries sorted lexicographically by name (Go docs). This matches systemd-stub's strcmp16 sort for ASCII filenames. The composite cmdline is built correctly: UKI.cmdline + " " + addon1.cmdline + " " + addon2.cmdline in sorted order.
```

crates/trident_api/src/constants.rs
```diff
@@ -165,6 +165,15 @@ pub const AB_VOLUME_A_NAME: &str = "A";
 /// A/B Volume B Name
 pub const AB_VOLUME_B_NAME: &str = "B";
 
+// ACL (Azure Container Linux) UKI disk layout defines fixed PARTUUIDs for the
+// USR A/B data partitions. These are from acl-scripts disk_layout_uki.json.
+
+/// ACL USR partition A PARTUUID
+pub const ACL_USR_A_PARTUUID: &str = "7130c94a-213a-4e5a-8e26-6cce9662f132";
```
frhuelsz: nit `uuid!(...)` macro to avoid runtime parse :)

Also, i feel tempted to put these in sysdefs::acl ? These are very high level constants that exist beyond Trident. (you can make the same argument about much of what lives here, but one step at a time?)

crates/trident/src/engine/newroot.rs
```diff
@@ -202,6 +220,39 @@ impl NewrootMount {
 
                 let fs_type = block_device.fstype.and_then(|fs_type| KernelFilesystemType::from(fs_type.as_str()).try_as_real());
 
+                // ACL-specific: if the staging device has a BTRFS filesystem UUID that
+                // collides with the active USR partition, bind-mount from the host's
+                // /usr instead. The verity-protected filesystem is read-only and the
+                // content is identical when UUIDs match, so the bind mount provides
+                // equivalent content for chroot provisioning.
+                if let Some(ref collision_uuid) = acl_collision_uuid {
+                    if *path == Path::new("/usr")
```
frhuelsz: nit: USR const

frhuelsz: (also i believe you can directly compare a path-like with `&'static str`)

bfjelds: d'oh!  thought i had fixed that. `<grumble>`

fintelia (dismissed):

tools/cmd/mkcosi/generator/generator.go
bfjelds: added comment to go code you referenced and updated pcrlock.rs to sort addons.

crates/trident/src/engine/boot/uki.rs
```diff
@@ -187,7 +310,9 @@ fn enumerate_trident_managed_ukis(
             .and_then(|filename| filename.strip_prefix("vmlinuz-"))
             .and_then(|f| f.split_once('-'))
             .filter(|(_, suffix)| {
-                suffix.contains("staged") || suffix.contains("azla") || suffix.contains("azlb")
+                let slot_a = uki_slot(AB_VOLUME_A_NAME);
+                let slot_b = uki_slot(AB_VOLUME_B_NAME);
+                suffix.contains("staged") || suffix.contains(&slot_a) || suffix.contains(&slot_b)
```
frhuelsz: nit: const for this string?

crates/trident/src/engine/ab_update.rs
bfjelds: Const usage fixed in fcf0b5e3 — now uses `USR_MOUNT_POINT_PATH` instead of a hardcoded `\"/usr\"` string.

Re the `Path::new` wrapper: `PathBuf` doesn't implement `PartialEq<str>`, so the `Path::new()` wrapping is needed here to compare `fs.mount_point` (a `PathBuf`) against the `&str` constant. Keeping as-is.

crates/trident/src/subsystems/storage/osimage.rs
bfjelds: Fixed in 7eeeec23 — moved mid-file imports to the top-of-file block in both osimage.rs and newroot.rs.

crates/trident/src/engine/acl.rs
bfjelds: Addressed in f8727a2c — the ACL-specific constants (PARTUUIDs, volume labels) were moved to `trident_api::constants`, and the `VerityRootHash` type was extracted to `osutils::verity_roothash`. The `acl.rs` file has been removed entirely; its logic was inlined or relocated.

crates/trident/src/engine/newroot.rs
bfjelds: Fixed in f1f6203a0 — now uses `block_devices::part_uuid_path()` instead of a hand-built format string.

crates/trident/src/engine/boot/uki.rs
bfjelds: Fixed in 53ec7713 — UKI slot names are now derived from the parent constants (`AZURE_LINUX_INSTALL_ID_PREFIX`, `AB_VOLUME_A_NAME`, `AB_VOLUME_B_NAME`) via a `uki_slot()` helper instead of hardcoding `azla`/`azlb`.

crates/trident/src/engine/acl.rs
bfjelds: Done in f8727a2c — introduced `osutils::verity_roothash::VerityRootHash` as a newtype wrapper around `String`. It implements `PartialEq` for comparing against other instances and provides `AsRef<str>`. The `acl.rs` module has been removed.

crates/trident/src/engine/newroot.rs
bfjelds: Good question. On Linux, `/dev/disk/by-uuid/` symlinks are created by udev from the filesystem UUID reported by `blkid`. If an ext4 and a BTRFS partition share the same UUID, both symlinks would target the same UUID path — last-created wins, making the link non-deterministic.

However, this code path uses `/dev/disk/by-partuuid/` (GPT partition UUIDs), not filesystem UUIDs. PARTUUIDs are always unique per partition regardless of filesystem type, so the ext4-vs-BTRFS collision scenario doesn't apply here.

The BTRFS UUID collision we handle is specifically between two BTRFS partitions (USR A and USR B) that share the same filesystem UUID because they were cloned from the same COSI image.

crates/trident/src/engine/boot/uki.rs
frhuelsz, line 381: this code may be doing something similar to the new function to cleanup, may be worth understanding what this was doing and how it interacts with the new code

crates/trident/src/engine/newroot.rs
bfjelds: Implemented in 2339336f. Changed return type from \Option<OsUuid>\ to \Result<Option<OsUuid>>\:

- \Ok(Some(uuid))\ — collision detected and verity-verified; use bind mount
- \Ok(None)\ — no collision (not ACL, not BTRFS, or different UUIDs)
- \Err(...)\ — collision detected but verification failed; clear error message instead of letting BTRFS produce a confusing kernel error

All four failure cases (no roothash, empty roothash, mismatch, unreadable /proc/cmdline) now bail with descriptive messages that include the colliding filesystem UUID.

crates/trident/src/engine/newroot.rs
bfjelds: Fixed in 5a165fad — now parses through `KernelFilesystemType::from().try_as_real()` and compares against `RealFilesystemType::Btrfs` instead of a raw string. This matches the pattern already used at line 221 in the same file.

crates/trident/src/engine/newroot.rs
bfjelds: Applied in 5a165fad — switched to `let Some(staging) = ... else { ... }` pattern.

crates/trident_api/src/constants.rs
bfjelds: Implemented in 3c2e7cb2 — created sysdefs::acl module with compile-time Uuid constants via uuid!() macro. Removed old string constants from trident_api::constants and updated all consumers (newroot.rs, osimage.rs, partitions.rs). No more runtime parsing.

crates/trident/src/engine/newroot.rs
bfjelds: Fixed in 5a165fad — now uses `USR_MOUNT_POINT_PATH` const.

Re comparing directly without `Path::new`: `PathBuf` doesn't implement `PartialEq<str>`, so the `Path::new()` wrapper is needed to compare `*path` (a `PathBuf`) against the `&str` constant. Keeping the wrapper.

crates/trident/src/subsystems/esp.rs
bfjelds: Deferring to Brian on this one — the ESP space management logic is tied to the A/B update flow which is the core of this PR. Taking it out would require restructuring the UKI staging path.

crates/trident/src/engine/boot/uki.rs
bfjelds: Fixed in 5a165fad — extracted UKI_FILENAME_PREFIX const. All three usages (both strip_prefix calls and the format! for new UKI filenames) now reference the const.

crates/trident/src/engine/newroot.rs
```diff
@@ -202,6 +220,39 @@ impl NewrootMount {
...
+                        do_bind_mount(active_usr, &target_path, MountFlags::RDONLY)
+                            .with_context(|| {
+                                format!(
+                                    "Failed to bind mount '{}' to '{}' \
+                                     for ACL BTRFS UUID collision workaround",
+                                    active_usr.display(),
+                                    target_path.display(),
+                                )
+                            })?;
+                        self.add_mount(target_path.clone());
```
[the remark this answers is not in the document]
bfjelds: Valid observation. On Linux, a single mount(2) call with MS_BIND|MS_RDONLY does not actually apply the read-only flag — the kernel ignores MS_RDONLY for the initial bind. A remount (MS_BIND|MS_REMOUNT|MS_RDONLY) is required to enforce it. However, in this specific case the source /usr is already mounted read-only via dm-verity, so the bind inherits the read-only property from the source. The MountFlags::RDONLY is defense-in-depth, not the primary protection. Adding a comment to clarify this would be useful — deferring to a follow-up.

crates/osutils/src/pcrlock.rs
```diff
@@ -750,8 +750,13 @@ where
             "UKI addon path '{}' exists, so generating .pcrlock file to measure it as well",
             uki_addon_path.display()
         );
-        for entry in fs::read_dir(&uki_addon_path).context("Failed to read UKI addons directory")? {
-            let entry = entry.context("Failed to get entry in UKI addons directory")?;
+        let mut addon_entries: Vec<_> = fs::read_dir(&uki_addon_path)
+            .context("Failed to read UKI addons directory")?
+            .filter_map(|e| e.ok())
+            .collect();
```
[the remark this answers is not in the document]
bfjelds: Fixed in 5a165fad — read_dir entry errors are now propagated via collect::<Result<Vec<_>, _>>().context() instead of being silently dropped by filter_map. A transient I/O error will now fail the pcrlock generation rather than producing incorrect measurements.

crates/trident/src/engine/newroot.rs
bfjelds: You were right — PathBuf can compare directly with &str. Fixed in eb2ea70a, dropped the Path::new wrapper.

crates/trident/src/engine/ab_update.rs
bfjelds: Correction to my earlier reply: Paco was right — PathBuf can compare directly with &str via Path's PartialEq<str> impl. Fixed the newroot.rs instance in eb2ea70a. The ab_update.rs comparison at line 72 could also drop Path::new, but that code predates this PR so leaving it alone.

crates/trident_api/src/config/host/storage/partitions.rs
```diff
@@ -181,12 +181,12 @@ pub enum PartitionType {
 impl PartitionType {
     /// Returns the ACL partition type that is treated as USR-like in storage validation.
     pub fn acl_usr() -> Self {
-        Self::Unknown(ACL_USR_PARTITION_TYPE_UUID)
+        Self::Unknown(sysdefs::acl::USR_PARTITION_TYPE_UUID)
```
frhuelsz: nit: import const directly :)

crates/trident/src/engine/boot/uki.rs
```diff
@@ -160,6 +197,98 @@ pub fn prepare_esp_for_uki(root_mount_point: &Path, esp_mount_path: &Path) -> Re
...
+    let target_suffix = format!("{target_slot}.efi");
+    let active_slot = active_slot_os_id(ctx);
+    let active_suffix = format!("{active_slot}.efi");
+    let trident_ukis = enumerate_trident_managed_ukis(&esp_uki_directory)?;
+
+    // 1. Remove trident-managed UKIs for the target slot+os-index
+    //    (all update indices, e.g. vmlinuz-100-azla0.efi, vmlinuz-102-azla0.efi).
+    for (_index, suffix, path) in &trident_ukis {
+        if *suffix == target_suffix {
+            debug!(
```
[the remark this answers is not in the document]
bfjelds: Invalid. The destructuring `for (_index, suffix, path) in &trident_ukis` yields `suffix: &String`, and `*suffix == target_suffix` compares `String == String` which compiles fine. Verified via `make check` (clippy -D warnings passes).

crates/trident/src/engine/boot/uki.rs
```diff
@@ -160,6 +197,98 @@ pub fn prepare_esp_for_uki(root_mount_point: &Path, esp_mount_path: &Path) -> Re
...
+    //      superseded as rollback).
+    //    In multiboot, install_index > 0 never owns the original UKI — it has
+    //    OS 0's partition refs baked in and can't boot other OS instances.
+    //    Note: this intentionally does NOT require both slots to be managed.
+    //    The ESP (~125 MB) cannot hold 3 UKIs (~50 MB each), so the original
+    //    must be removed before staging to free space. Staging already breaks
+    //    rollback to the original, so deleting it here is acceptable.
+    let has_active_slot_uki = trident_ukis
+        .iter()
+        .any(|(_index, suffix, _)| *suffix == active_suffix);
```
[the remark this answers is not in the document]
bfjelds: Same as above — this compiles correctly. `*suffix` derefs `&String` to `String` and the comparison works. `make check` passes clean.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -257,6 +272,120 @@ fn validate_filesystem_uniqueness(
...
+
+    debug!(
+        "ACL duplicate FS UUID '{}' on /usr is safe: verity root hashes match ({}...)",
+        uuid_str,
+        staging_hash.preview()
+    );
+
+    // Optional: If COSI partition metadata is available, validate that the staging
+    // USR partition PARTUUID matches a known ACL USR slot.
+    if let Some(mut partitions) = os_image.partitions() {
```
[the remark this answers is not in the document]
bfjelds: Invalid. `mut` is required here because `Iterator::any()` takes `&mut self`. Without `mut`, the call would fail to compile. `make check` passes with no `unused_mut` warning.

crates/trident/src/engine/newroot.rs
```diff
@@ -202,6 +222,39 @@ impl NewrootMount {
...
+                // ACL-specific: if the staging device has a BTRFS filesystem UUID that
+                // collides with the active USR partition, bind-mount from the host's
+                // /usr instead. The verity-protected filesystem is read-only and the
+                // content is identical when UUIDs match, so the bind mount provides
+                // equivalent content for chroot provisioning.
+                if let Some(ref collision_uuid) = acl_collision_uuid {
+                    if *path == USR_MOUNT_POINT_PATH
+                        && fs_type == Some(RealFilesystemType::Btrfs)
+                        && block_device.fsuuid.as_ref() == Some(collision_uuid)
+                    {
```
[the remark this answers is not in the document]
bfjelds: Invalid. `PathBuf` implements `PartialEq<&str>` (via `Path`'s blanket impls through `AsRef<OsStr>`), so `*path == USR_MOUNT_POINT_PATH` compiles and works correctly. Verified via `make check`.

crates/trident_api/src/config/host/storage/partitions.rs
bfjelds: Fixed in 5a87bac4 — imported `USR_PARTITION_TYPE_UUID` directly from `sysdefs::acl`.

crates/trident/src/engine/boot/uki.rs
bfjelds: Good catch. Fixed in 3260b650 — the inline cleanup in `update_uki_boot_files` now delegates to `remove_uki_and_addons` instead of duplicating the delete logic. This makes both cleanup paths consistent (NotFound-safe) and the inline delete is now an idempotent safety net since `cleanup_ukis_before_staging` should have already removed the files.

bfjelds: ooooh!  thanks for finding that :)

will work on un-duplicating.

crates/trident/src/engine/newroot.rs
```diff
@@ -339,6 +392,121 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
...
+
+    let active_path = block_devices::part_uuid_path(active_partuuid);
+    let update_path = block_devices::part_uuid_path(update_partuuid);
+
+    let Ok(active_dev) = lsblk::get(&active_path) else {
+        return Ok(None);
+    };
+    let Ok(update_dev) = lsblk::get(&update_path) else {
+        return Ok(None);
+    };
```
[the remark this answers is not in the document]
bfjelds: Fixed in 5327d8e8 — switched to lsblk::try_get which returns Ok(None) for missing devices (expected on non-ACL systems) but propagates real lsblk execution/parse errors.

crates/trident/src/subsystems/storage/osimage.rs
```diff
@@ -257,6 +272,117 @@ fn validate_filesystem_uniqueness(
...
+            return Err(TridentError::new(
+                InvalidInputError::DuplicateFsUuidAclVerificationFailed {
+                    uuid: uuid_str,
+                    mount_point: mount_str.to_string(),
+                    reason: "staging /usr image has no verity root hash in COSI metadata"
+                        .to_string(),
+                },
+            ));
+        }
+    };
```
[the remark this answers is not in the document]
bfjelds: Fixed in 5327d8e8 — split into two checks: first verifies a /usr filesystem exists in COSI metadata, then separately checks for a verity root hash. Error messages now distinguish 'no /usr filesystem found in COSI image metadata' from '/usr filesystem has no verity root hash'.

crates/trident/src/subsystems/esp.rs
bfjelds: i'm not sure there is a point to this PR without this.  not sure why it fits in 25 more than 24.

bfjelds: the code in this PR can be validated using https://dev.azure.com/mariner-org/ACL/_git/acl-scripts/pullrequest/27615 and https://dev.azure.com/mariner-org/ACL/_git/acl-pipelines/pullrequest/27673.  if we take this out, then we are merging a bunch of code that can't be verified.

crates/trident/src/engine/newroot.rs
```diff
@@ -202,6 +222,39 @@ impl NewrootMount {
 
                 let fs_type = block_device.fstype.and_then(|fs_type| KernelFilesystemType::from(fs_type.as_str()).try_as_real());
 
+                // ACL-specific: if the staging device has a BTRFS filesystem UUID that
+                // collides with the active USR partition, bind-mount from the host's
+                // /usr instead. The verity-protected filesystem is read-only and the
+                // content is identical when UUIDs match, so the bind mount provides
+                // equivalent content for chroot provisioning.
+                if let Some(ref collision_uuid) = acl_collision_uuid {
+                    if *path == USR_MOUNT_POINT_PATH
```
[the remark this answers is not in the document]
bfjelds: Invalid — the code already uses Path::new(USR_MOUNT_POINT_PATH) as of 9e0ee5a4. This comment is based on stale code.

crates/trident/src/engine/newroot.rs
```diff
@@ -339,6 +392,125 @@ fn should_be_bind_mounted(fs_type: Option<RealFilesystemType>) -> bool {
...
+    let Some(active_dev) =
+        lsblk::try_get(&active_path).context("Failed to query active ACL USR partition")?
+    else {
+        return Ok(None);
+    };
+    let Some(update_dev) =
+        lsblk::try_get(&update_path).context("Failed to query update ACL USR partition")?
+    else {
+        return Ok(None);
+    };
```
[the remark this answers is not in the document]
bfjelds: Valid. Fixed in 0a6dee77 — added path existence check before calling lsblk::try_get. On non-ACL systems the PARTUUID symlinks don't exist, so we return Ok(None) early. Real lsblk failures on ACL systems (where the paths do exist) still propagate as errors.

tools/cmd/mkcosi/generator/generator.go
```diff
@@ -1085,17 +1086,73 @@ func findUkiEntries(espMountPath string, espMountPoint string) []metadata.System
 			continue
 		}
```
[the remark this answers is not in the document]
bfjelds: Invalid. The check requires BOTH .cmdline AND .uname to be empty before skipping. A valid UKI always has at least .uname. A file with neither section is not a UKI, so scanning its addon directory would be meaningless — addons supplement a UKI, they don't replace one.

crates/trident/src/engine/boot/uki.rs
```diff
@@ -160,6 +197,98 @@ pub fn prepare_esp_for_uki(root_mount_point: &Path, esp_mount_path: &Path) -> Re
...
+                addon_dir.display()
+            )
+        })?;
+    }
+    Ok(())
+}
+
+/// Cleans up old UKIs for the target slot before staging a new one, freeing
+/// ESP space.  This runs for all UKI-based A/B updates (not just ACL) because
+/// ESP space constraints are a general concern.
```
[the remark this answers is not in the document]
bfjelds: Invalid. The function is a safe no-op on clean installs: enumerate_trident_managed_ukis returns an empty list (no existing UKIs), and has_active_slot_uki is false, so neither cleanup branch executes. The doc describes the function's purpose (ESP space management for A/B updates), not a precondition. Adding an explicit A/B-only guard would be unnecessary defensive code.

fintelia (approved):

crates/trident/src/engine/newroot.rs
```diff
@@ -202,6 +222,39 @@ impl NewrootMount {
...
+                // collides with the active USR partition, bind-mount from the host's
+                // /usr instead. The verity-protected filesystem is read-only and the
+                // content is identical when UUIDs match, so the bind mount provides
+                // equivalent content for chroot provisioning.
+                if let Some(ref collision_uuid) = acl_collision_uuid {
+                    if *path == Path::new(USR_MOUNT_POINT_PATH)
+                        && fs_type == Some(RealFilesystemType::Btrfs)
+                        && block_device.fsuuid.as_ref() == Some(collision_uuid)
+                    {
+                        let active_usr = Path::new("/usr");
```
frhuelsz, line 235: nit nit nit: use of literal `/usr`, prefer the constant

crates/trident_api/src/config/host/storage/partitions.rs
```diff
@@ -181,12 +178,12 @@ pub enum PartitionType {
 impl PartitionType {
     /// Returns the ACL partition type that is treated as USR-like in storage validation.
     pub fn acl_usr() -> Self {
-        Self::Unknown(ACL_USR_PARTITION_TYPE_UUID)
+        Self::Unknown(USR_PARTITION_TYPE_UUID)
```
frhuelsz, line 181: super nit but i just noticed that this is weird without context because it looks like a generic "USR: thing, only when you check the import you see it's coming from ACL, perhaps the consta should be named ACL_... (albeit a bit redundant) so that it's easily identifyable anywhere?

frhuelsz (approved):

crates/trident_api/src/config/host/storage/partitions.rs
bfjelds: will address in followup

crates/trident/src/engine/newroot.rs
bfjelds: will address in followup

crates/trident_api/src/config/host/storage/partitions.rs
bfjelds: https://github.com/microsoft/trident/pull/687

crates/trident/src/engine/newroot.rs
bfjelds: https://github.com/microsoft/trident/pull/687

PR #685, opened 2026-06-11
docs: Enhance copilot-instructions.md

frhuelsz:
# 🔍 Description
 
Add a bunch of info on repo structure and code conventions for both writer and reviewer agents.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,6 +1,13 @@
-# Copilot PR Review Instructions
+# Copilot Agent Instructions
 
-## Scope
+These instructions apply to Copilot agents working in this repo — both agents
+writing code and agents reviewing pull requests. The PR-review rules below are
+review-only. The "Nits" and "Architecture & structural soundness" sections
+apply to **both** writers (follow them when generating new code) and
+reviewers (see [Reviewer etiquette](#reviewer-etiquette-nits--architecture)
+for when to surface them).
+
+## PR Review Scope
 - Only comment on issues that are **specific to the diff** (avoid generic best-practice reminders).
 - Avoid repeating the same point across multiple files. If one example demonstrates a pattern, mention it once and reference the pattern.
 - **ALWAYS check previous reviews** before commenting. Do NOT repeat points that have already been made in previous reviews if they have been acknowledged, dismissed, or closed.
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -1,26 +1,307 @@
...
+## Nits (Rust, applies to writers and reviewers)
+
+Trident formats with `rustfmt` (workspace-wide — see `rustfmt.toml`).
+Everything below is in addition to `cargo fmt` and is **aspirational**: real
+files don't follow every rule 100% of the time. A counter-example in the tree
+is not a license to ignore the rule in new code.
+
+### Imports
+
+1. **Order groups, separated by a single blank line:** (1) `std`, (2) external
```
bfjelds, line 48: do we want to advise that we prefer at max a single `::` in code when referencing things from using? (so no code appearances of foo::bar::blah::Something)

frhuelsz: That is okay, but doesn't happen much. If we're genuinely _only_ using `Something` from crate `foo`, then there is really no way around it. But good point about nesting imports as much as possible. I'll add that.

.github/copilot-instructions.md
```diff
@@ -1,26 +1,307 @@
...
+
+### When in doubt
+
+Prefer **adding a small wrapper to the right layer and calling it** over
+duplicating a system invocation in a subsystem. The two most common
+review nits in this category are: (a) "this `std::process::Command::new("foo")`
+belongs in `osutils::foo`, routed through `Dependency::Foo`"; (b) "this
+constant duplicates `trident_api::constants::FOO`".
+
+## Reviewer etiquette (Nits & Architecture)
```
bfjelds, line 286: do you think we can tell Copilot to avoid repeating comments?

.github/copilot-instructions.md
bfjelds: it doesn't seem to be uncommon for ai generated code

bfjelds (approved):

.github/copilot-instructions.md
```diff
@@ -1,26 +1,306 @@
...
+Everything below is in addition to `cargo fmt` and is **aspirational**: real
+files don't follow every rule 100% of the time. A counter-example in the tree
+is not a license to ignore the rule in new code.
+
+### Imports
+
+1. **Order groups, separated by a single blank line:** (1) `std`, (2) external
+   crates, (3) other workspace crates (e.g. `osutils`, `sysdefs`, `trident_api`,
+   `trident-proto`), (4) `crate::`, (5) `super::`. Inside a single group, keep
+   `use` lines alphabetical by path.
```
fintelia, line 51: We could get closer with [this rustfmt option](https://rust-lang.github.io/rustfmt/?version=v1.9.0&search=import#StdExternalCrate%5C%3A) though it isn't quite what we want

.github/copilot-instructions.md
frhuelsz: i think each review is made by a new independent agent that only has access to the repo. This has been in here for a while:
`- **ALWAYS check previous reviews** before commenting. Do NOT repeat points that have already been made in previous reviews if they have been acknowledged, dismissed, or closed.`

.github/copilot-instructions.md
frhuelsz: clarified online, we were talking about different things, both are now considered

PR #718, opened 2026-07-27
ACL: firstboot UKI addon only applies for clean install

bfjelds:
# 🔍 Description

ACL AB-update COSI payloads are typically built from a VHD that may (for Azure) be configured via <uki>.efi.extra.d/firstboot.addon.efi to treat the next boot as the first boot. COSI files can be used for install or update, so trident must choose when to apply firstboot.addon.efi in the target OS according to whether install or update is invoked.

`enforce_firstboot_addon_policy(...)`: 
* ACL-specific
* on a clean install, leaves the addon as staged (or copies it from the acl/uki-addons/ template if only present there)
* on any other servicing type, removes it from the staged addon directory if present

Trident validation: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=1169144&view=results
ACL validation: https://dev.azure.com/mariner-org/ACL/_build/results?buildId=1169180&view=results

### DIFF
--- a/crates/trident/src/engine/boot/uki.rs
+++ b/crates/trident/src/engine/boot/uki.rs
@@ -655,16 +655,15 @@ pub(crate) fn enforce_firstboot_addon_policy(
     esp_mount_path: &Path,
     servicing_type: ServicingType,
 ) -> Result<(), Error> {
-    let is_clean_install = match servicing_type {
-        ServicingType::CleanInstall => true,
-        ServicingType::AbUpdate => false,
-        other => {
-            return Err(anyhow!(
-                "enforce_firstboot_addon_policy called with unexpected servicing type \
-                 {other:?}; expected CleanInstall or AbUpdate"
-            ))
-        }
-    };
+    ensure!(
+        matches!(
+            servicing_type,
+            ServicingType::CleanInstall | ServicingType::AbUpdate
+        ),
+        "enforce_firstboot_addon_policy called with unexpected servicing type \
+         {servicing_type:?}; expected CleanInstall or AbUpdate"
+    );
+    let is_clean_install = servicing_type == ServicingType::CleanInstall;
 
     let staging_addon_dir = join_relative(mount_point, esp_mount_path)
         .join(UKI_DIRECTORY)
--- a/crates/trident/src/engine/boot/uki.rs
+++ b/crates/trident/src/engine/boot/uki.rs
@@ -639,7 +639,7 @@ const FIRSTBOOT_ADDON_FILENAME: &str = "firstboot.addon.efi";
 /// This is ACL-specific; callers gate on `ctx.image_distro().is_acl()`,
 /// matching the verity addon's existing convention — non-ACL images never
 /// reach this function.
-pub fn enforce_firstboot_addon_policy(
+pub(crate) fn enforce_firstboot_addon_policy(
     image_esp_mount: &Path,
     mount_point: &Path,
     esp_mount_path: &Path,
--- a/crates/trident/src/engine/boot/uki.rs
+++ b/crates/trident/src/engine/boot/uki.rs
@@ -1562,6 +1562,14 @@ mod tests {
             err.to_string().contains("unexpected servicing type"),
             "Expected 'unexpected servicing type' error, got: {err}"
         );
+        // Confirms the servicing type's actual Debug value is interpolated
+        // via the captured identifier ({servicing_type:?}), not left as a
+        // literal placeholder — Rust supports captured identifiers in
+        // format/ensure! strings since 1.58.
+        assert!(
+            err.to_string().contains("RuntimeUpdate"),
+            "Expected the actual servicing type value in the error, got: {err}"
+        );
     }
 
     /// Helper: creates an ESP directory structure with UKI files and returns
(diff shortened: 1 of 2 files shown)


crates/trident/src/subsystems/esp.rs
```diff
@@ -311,6 +311,22 @@ fn copy_file_artifacts(
                     &ctx.esp_mount_path,
                     target_volume,
                 )?;
+
+                // ACL images ship the first-boot addon pre-populated in the
+                // image's live .extra.d/ (already staged verbatim above by
+                // stage_uki_on_esp) so a genuine first boot works out of the
+                // box. But every ACL AB-update COSI is built from a
+                // never-booted image, so that addon is still present even
+                // when this is not actually a first boot. Enforce it's only
```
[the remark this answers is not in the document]
bfjelds: Great comment — fixed in 4babe035 ("Enforce that it's only ever staged...").

crates/trident/src/engine/boot/uki.rs
```diff
@@ -604,6 +604,115 @@ pub fn activate_verity_addon_for_target_volume(
...
+/// `.extra.d/` directory (already staged by `stage_uki_on_esp` before this
+/// function runs) still carries this addon regardless of whether this is
+/// truly a first boot. Left unremoved, it reintroduces
+/// `flatcar.first_boot=detected` on every update, hanging
+/// `ignition-fetch.service` after the post-update reboot.
+///
+/// This is ACL-specific; callers gate on `ctx.image_distro().is_acl()`,
+/// matching the verity addon's existing convention — non-ACL images never
+/// reach this function.
+pub fn enforce_firstboot_addon_policy(
```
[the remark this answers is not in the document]
bfjelds: Great comment — narrowed to pub(crate) in 4babe035; it's only ever called from subsystems/esp.rs within this crate.

crates/trident/src/engine/boot/uki.rs
```diff
@@ -604,6 +604,115 @@ pub fn activate_verity_addon_for_target_volume(
...
+            // image's live .extra.d/ directory. Nothing further to do.
+            trace!(
+                "'{FIRSTBOOT_ADDON_FILENAME}' already staged from the image's live addon \
+                 directory; leaving as-is for clean install"
+            );
+            return Ok(());
+        }
+
+        let template_path = image_esp_mount
+            .join(VERITY_ADDON_TEMPLATES_DIR)
```
fintelia: We should probably rename this constant now

fintelia (dismissed):

crates/trident/src/engine/boot/uki.rs
bfjelds: good call

fintelia (dismissed):

crates/trident/src/subsystems/esp.rs
```diff
@@ -311,6 +311,22 @@ fn copy_file_artifacts(
...
+                // never-booted image, so that addon is still present even
+                // when this is not actually a first boot. Enforce that it's
+                // only ever staged for a clean install; strip it otherwise
+                // so Ignition doesn't re-trigger and hang on the post-update
+                // reboot.
+                uki::enforce_firstboot_addon_policy(
+                    temp_mount_dir,
+                    mount_point,
+                    &ctx.esp_mount_path,
+                    ctx.servicing_type == ServicingType::CleanInstall,
```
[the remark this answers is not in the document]
bfjelds: ACL with trident is always ab-update.  If this were invoked somehow without ab-update (install), then the contents of extra.d would be preserved.

frhuelsz: silly nit, could we pass the servicing type directly instead of a bool? That also allows us to error our on unexpected servicing types.

crates/trident/src/engine/boot/uki.rs
bfjelds: done

crates/trident/src/subsystems/esp.rs
bfjelds: done

crates/trident/src/engine/boot/uki.rs
```diff
@@ -604,6 +606,133 @@ pub fn activate_verity_addon_for_target_volume(
...
+    servicing_type: ServicingType,
+) -> Result<(), Error> {
+    ensure!(
+        matches!(
+            servicing_type,
+            ServicingType::CleanInstall | ServicingType::AbUpdate
+        ),
+        "enforce_firstboot_addon_policy called with unexpected servicing type \
+         {servicing_type:?}; expected CleanInstall or AbUpdate"
+    );
```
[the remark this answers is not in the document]
bfjelds: Great comment — this actually already works correctly. Rust has supported captured identifiers in format strings (including inside `anyhow::ensure!`/`format!`) since 1.58, so `{servicing_type:?}` inside the string literal does interpolate the live variable, not a literal placeholder. This is the same pattern already used elsewhere in this file (e.g. `{FIRSTBOOT_ADDON_FILENAME}`).

To make this verifiable rather than just asserted, I added a test assertion in 0d195b63 confirming the error message for `ServicingType::RuntimeUpdate` actually contains the string `RuntimeUpdate` (not a literal `{servicing_type:?}`).

frhuelsz (approved):

PR #729, opened 2026-08-05
trident: promote rollback to stable v1; adjust rollback return type to detect no-servicing

bfjelds:
## Summary

Promotes manual-rollback's `Rollback`/`RollbackStage`/`RollbackFinalize` from `trident.v1preview.RollbackService` to the stable `trident.v1.RollbackService`.

Align manual rollback's no-op reporting with `update()`/`install()`: a no-op rollback (empty rollback chain, or host not in a rollback-eligible state) now reports `(ExitKind::Done, ServicingType::NoActiveServicing)` instead of a bare `ExitKind::Done`.

## Context

This is the first step in enabling trident-acl-agent to run updates and rollbacks.  Related PRs:
* [this PR] https://github.com/microsoft/trident/pull/729
* https://github.com/microsoft/trident/pull/733
* https://github.com/microsoft/trident/pull/730
* https://github.com/microsoft/trident/pull/731

## Validation

* Just this change (checking for regressions): https://dev.azure.com/mariner-org/ECF/_build/results?buildId=1176186&view=results
* Validate grpc invocation with full trident-acl-agent tests: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=1176259&view=results

### DIFF
--- a/crates/trident/src/server/tridentserver/services/rollback.rs
+++ b/crates/trident/src/server/tridentserver/services/rollback.rs
@@ -59,6 +59,9 @@ impl RollbackService for TridentServer {
         request: Request<RollbackRequest>,
     ) -> Result<Response<Self::RollbackStream>, Status> {
         let req = request.into_inner();
+        let Some(stage) = req.stage else {
+            return Err(Status::invalid_argument("Missing stage configuration"));
+        };
         let Some(finalize) = req.finalize else {
             return Err(Status::invalid_argument("Missing finalize configuration"));
         };
@@ -79,7 +82,7 @@ impl RollbackService for TridentServer {
                     .message("Failed to open datastore")?;
 
                 let (invoke_if_next_is_runtime, invoke_available_ab) =
-                    manual_rollback_flags(req.kind())?;
+                    manual_rollback_flags(stage.kind())?;
 
                 trident
                     .rollback(
--- a/crates/trident/src/server/tridentserver/services/rollback.rs
+++ b/crates/trident/src/server/tridentserver/services/rollback.rs
@@ -1,65 +1,217 @@
 use tonic::{async_trait, Request, Response, Status};
 
-use trident_api::error::{InternalError, TridentError};
-use trident_proto::v1preview::{
-    rollback_service_server::RollbackService, CheckRollbackRequest, CheckRollbackResponse,
-    GetRollbackChainRequest, GetRollbackChainResponse, GetRollbackTargetRequest,
-    GetRollbackTargetResponse, RollbackFinalizeRequest, RollbackRequest, RollbackStageRequest,
+use trident_api::{
+    config::{Operation, Operations},
+    error::{TridentError, TridentResultExt},
+};
+use trident_proto::v1::{
+    rollback_service_server::RollbackService, ManualRollbackKind, RollbackFinalizeRequest,
+    RollbackRequest, RollbackStageRequest,
 };
 
-use crate::server::{
-    tridentserver::{RebootDecision, ServicingResponseStream},
-    TridentServer,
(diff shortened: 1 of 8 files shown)


crates/trident/src/server/tridentserver/services/rollback.rs
```diff
@@ -1,65 +1,217 @@
...
+                let mut trident: Trident =
+                    Trident::new(None, &data_store_path, logstream, tracestream)
+                        .message("Failed to initialize Trident")?;
+
+                let mut datastore = DataStore::open_or_create(&data_store_path)
+                    .message("Failed to open datastore")?;
+
+                let (invoke_if_next_is_runtime, invoke_available_ab) =
+                    manual_rollback_flags(req.kind())?;
+
```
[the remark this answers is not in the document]
bfjelds: Great catch to flag — I checked this carefully since it's a real compile-risk pattern in general, but it turns out `req` isn't actually left partially moved here.

`RollbackFinalizeRequest` (the type of `req.finalize`) derives `Copy` in the generated proto code, since its only field, `reboot: Option<RebootManagement>`, is itself `Copy`:

```rust
#[derive(Clone, Copy, PartialEq, Eq, Hash, ::prost::Message)]
pub struct RollbackFinalizeRequest {
    pub reboot: ::core::option::Option<RebootManagement>,
}
```

Since the field type is `Copy`, `let Some(finalize) = req.finalize` copies the value out rather than moving it, so `req` stays fully intact and `req.kind()` afterward is valid. I confirmed this two ways:

1. `cargo build -p trident`, `cargo clippy --tests --all-features -- -D warnings`, and `cargo test` all pass cleanly on this exact code.
2. An isolated repro: the same pattern with a non-`Copy` struct does hit E0382 as expected, but with a `Copy`-deriving struct matching `RollbackFinalizeRequest`'s actual shape, it compiles clean — confirming the field's `Copy`-ness is what makes this safe.

No change needed here.

proto/trident/v1/rollback_service.proto
```diff
@@ -0,0 +1,57 @@
...
+//
+// This is the stable (v1) subset of the rollback contract: Rollback,
+// RollbackStage, and RollbackFinalize, the operations trident-acl-agent
+// depends on for AB-update rollback. CheckRollback, GetRollbackChain, and
+// GetRollbackTarget remain in trident.v1preview (see rollback_service.proto
+// there). CheckRollback was briefly promoted here but was demoted back to
+// preview once RollbackStage/RollbackFinalize/Rollback started reporting
+// ServicingKind::NoneRequired for a no-op (same as every other servicing
+// RPC), which removed trident-acl-agent's only reason to call it - see
+// ServicingResponse.servicing_kind in servicing.proto.
```
bfjelds, line 11: super nit: excessive commenting here, first line is probably sufficient.

frhuelsz (dismissed):

proto/trident/v1/rollback_service.proto
```diff
@@ -0,0 +1,57 @@
...
+  ANY_ROLLBACK_REQUESTED = 1;
+  // Rollback of an A/B update was requested
+  AB_ROLLBACK_REQUESTED = 2;
+  // Rollback of a runtime update was requested
+  RUNTIME_ROLLBACK_REQUESTED = 3;
+}
+
+message RollbackRequest {
+  // The rollback kind parameter for the servicing operation.
+  ManualRollbackKind kind = 1;
```
frhuelsz: prefer embedding the RollbackStageRequest here

frhuelsz (dismissed):

frhuelsz (approved):

PR #730, opened 2026-08-05
trident-acl-agent: rewrite as annotation-based ACL update agent

bfjelds:
## Summary

* `trident-acl-agent` implements annotation-based design (https://msazure.visualstudio.com/One/_git/Compute-ACL-Update-Service?version=GCeb7e534b2415ad52b37ef22fd49685e81e56c8aa&path=/docs/update-trigger-design.md): watching Node's `<prefix>/update-request` annotation, driving update stage/finalize, rollback stage/finalize, and commit gRPC operations against tridentd, and writing `<prefix>/update-status` and `<prefix>/update-commit-status`.
* includes CLI option `--validate-connection` for kubernetes|tridentd|nebraska.
* Adds docs/Explanation/Trident-ACL-Agent.md
* Updates trident.spec to install new unit file and publish trident-acl-agent in azurelinux builds.

## Context

This is the second step in enabling trident-acl-agent to run updates and rollbacks.  Related PRs:
* https://github.com/microsoft/trident/pull/729
* https://github.com/microsoft/trident/pull/733 (via cherry-pick, not actually stacked)
* [this PR] https://github.com/microsoft/trident/pull/730
* https://github.com/microsoft/trident/pull/731

## Validation

* This PR: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=1186399&view=results
* Validate trident-acl-agent implementation with full trident-acl-agent tests: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=1186400&view=results

PR #733, opened 2026-08-06
feat(acl-agent): nebraska client module (Omaha protocol, invariant-encoded)

frhuelsz:
## Summary

Adds a self-contained **`nebraska`** module in `crates/trident-acl-agent/src/nebraska/` implementing the Nebraska/Omaha update protocol. Scoped strictly to the protocol — no Trident gRPC, reboot, commit, or CLI logic — so it is reusable by the current agent and a future TAA.

The design goal is to **make illegal protocol states unrepresentable**. Several Nebraska behaviours fail *silently* (a client can look fine over the wire while wedging the fleet's state); this module encodes those invariants in the type system. Behavioural reference: `knowledge/topics/nebraska-client-protocol.md` (pacobot repo), cited throughout the rustdoc.

## Public API

- `Client<T = ReqwestTransport>` — bound to one app + track + `MachineId`; methods `check_for_update`, `report_progress`, `complete_after_reboot`, `report_failure`.
- `CheckOutcome` = `UpToDate | UpdateAvailable(UpdateOffer) | UpdateInProgress`.
- `UpdateOffer { version: semver::Version, package_url: Url }`.
- `ProgressEvent` = `DownloadStarted | DownloadFinished | Installed` (the only publicly constructible events).
- `MachineId` (validated, unbraced), `AppStatus`/`UpdateCheckStatus` (with `Other` catch-all), `Transport`/`ReqwestTransport`, `NebraskaError`.

## Invariants encoded

- Only the six whitelisted `(eventtype,eventresult)` pairs are constructible — no raw integers in the public API.
- `track` is a `Client` field → impossible to omit on any request (incl. event-only).
- `error-updateInProgressOnInstance` is a first-class `CheckOutcome`, not an error; unknown status strings map to `Other` rather than failing to parse.
- `MachineId` rejects braced ids; `from_uuid` is unbraced.
- Versions are `semver::Version`, not `String`.
- Terminal events are emitted only via named methods; completion is the batched `3/2 + ping + updatecheck` request.

### Design note: all-or-nothing (invariant #2)

Chosen a **documented plain API with a typed progress/terminal split**, not a typestate/RAII guard: the terminal event fires *after a reboot*, in a different process, so no in-process typestate can span the commitment, and an RAII guard would fire at the wrong moment (at reboot). Rationale in `nebraska/README.md`.

## Tests (hermetic — mock `Transport`, no network)

- Wire-format XML for update-check, each progress event, and the batched completion.
- Response parsing: update offer (incl. empty `<actions>`), `noupdate`, `error-updateInProgressOnInstance` + `error-internal`, and an unknown status (must not panic).
- Whitelist coverage: the event enums only emit whitelisted pairs.
- `MachineId` unbraced invariant (incl. rejecting a braced input) and empty rejection.
- semver handling incl. rejection of non-semver.
- Client behaviour: offer with joined package URL, no-update, in-progress mapping, wrong-app-id error, and that progress/terminal requests carry `track` + the right event.

24 module unit tests + 1 doctest; the crate's existing 26 tests still pass. `cargo fmt` / `cargo clippy --all-targets -- -D warnings` clean.

## Notes

- Exposes the crate as a library (new `lib.rs`) so the module is importable.
- The existing agent + ad-hoc `omaha` module are left **unchanged**; adoption path is described in `nebraska/README.md`. Migrating the agent is a deliberate follow-up.


crates/trident-acl-agent/src/omaha/mod.rs
```diff
@@ -1,5 +1,23 @@
 //! Super basic implementation of the Omaha protocol as defined in
 //! https://github.com/google/omaha/blob/main/doc/ServerProtocol.md.
+//!
```
bfjelds, line 3: should omaha code be removed?

frhuelsz: In a follow up, I didn't want to break everything at once.

crates/trident-acl-agent/src/nebraska/README.md
```diff
@@ -0,0 +1,120 @@
+# `nebraska` client module
+
+A self-contained Rust client for the [Nebraska](https://github.com/flatcar/nebraska)
```
bfjelds, line 3: should we call out a specific version (commit?) of Nebraska that this implements?

crates/trident-acl-agent/src/nebraska/wire.rs
```diff
@@ -0,0 +1,509 @@
...
+    #[serde(rename = "app")]
+    app: App,
+}
+
+impl Request {
+    /// Builds a request carrying a single app.
+    pub(super) fn new(app: App) -> Self {
+        Self {
+            protocol: OMAHA_PROTOCOL,
+            version: env!("CARGO_PKG_VERSION"),
```
bfjelds: should this be something like what trident uses:

``` rust
pub const TRIDENT_VERSION: &str = match option_env!("TRIDENT_VERSION") {
    Some(v) => v,
    None => env!("CARGO_PKG_VERSION"),
};
```

crates/trident-acl-agent/src/nebraska/client.rs
```diff
@@ -0,0 +1,815 @@
...
+    /// instance it has never seen. Emitting a progress event commits the caller
+    /// to eventually reporting a terminal event — see the [type docs](Client).
+    pub fn report_progress(
+        &self,
+        current_version: &Version,
+        event: ProgressEvent,
+    ) -> Result<(), NebraskaError> {
+        let app = self.app(current_version).with_event(event.wire());
+        let response = self.send(app)?;
+        self.require_app_present(&response)?;
```
bfjelds: Thanks for the clear invariant-focused design here — really nice work modeling the whitelisted event pairs and mandatory `track`.

One thing I'd love a second look at: `report_progress` and `report_failure` both call `require_app_present(&response)?` and then return `Ok(())`, but `require_app_present` only checks that the app id is present (`self.app_response(response).map(|_| ())`) — it doesn't inspect `app.status`. Only `interpret_check` (used by `check_for_update`/`complete_after_reboot`) validates status.

Would it make sense for these two methods to also check `app.status` before returning success? As written, if Nebraska responds with an error status on an HTTP 200 for one of these calls, the caller would see `Ok(())` even though the event wasn't accepted — which seems like exactly the kind of silent-failure case this module is designed to prevent, especially on the `report_failure` recovery path.

crates/trident-acl-agent/src/nebraska/client.rs
```diff
@@ -0,0 +1,815 @@
...
+/// hit. Blocks the current thread between attempts.
+fn retry<T>(
+    policy: RetryPolicy,
+    mut op: impl FnMut() -> Result<T, NebraskaError>,
+) -> Result<T, NebraskaError> {
+    let mut backoff = policy.initial_backoff;
+    let mut attempt = 1;
+    loop {
+        match op() {
+            Ok(value) => return Ok(value),
```
bfjelds, line 136: Really appreciate the comment above `try_complete_after_reboot` explaining that a still-in-progress status means the caller should retry — that's a great callout of a subtle protocol behavior.

Looking at `retry()`, though, it returns immediately on any `Ok(value)`, including `Ok(CheckOutcome::UpdateInProgress)`. Since `complete_after_reboot` wraps `try_complete_after_reboot` in this `retry()`, it looks like the "still in progress" outcome wouldn't actually trigger another attempt today. Could `retry()` (or the completion path) be adjusted to treat `UpdateInProgress` as a retryable outcome rather than a terminal success? Want to flag this since it seems to affect the module's main safety guarantee around the post-reboot completion event.

crates/trident-acl-agent/src/nebraska/transport.rs
```diff
@@ -0,0 +1,71 @@
...
+/// (connection failures) or [`NebraskaError::Http`] (non-success status).
+pub trait Transport {
+    /// POSTs `body` to `endpoint` and returns the response body as a string.
+    fn post_xml(&self, endpoint: &Url, body: &[u8]) -> Result<String, NebraskaError>;
+}
+
+/// The default [`Transport`], backed by a blocking `reqwest` client.
+#[derive(Debug, Default)]
+pub struct ReqwestTransport {
+    client: reqwest::blocking::Client,
```
bfjelds, line 53: Great job giving callers a clean `Transport` seam here for testing.

One question: `ReqwestTransport::default()` builds a plain `reqwest::blocking::Client` without an explicit timeout. Since `complete_after_reboot`'s docs promise a bounded retry window, could a hung DNS/TCP/TLS attempt block `.send()` long enough that the retry loop never gets a chance to back off and retry? Would it be worth adding a default timeout to the transport (or updating the docs to clarify what "bounded" means in that case)?

crates/trident-acl-agent/src/nebraska/client.rs
```diff
@@ -0,0 +1,815 @@
...
+        )
+    }
+
+    /// Serializes, sends, and parses a request/response round-trip.
+    fn send(&self, app: App) -> Result<wire::Response, NebraskaError> {
+        let request = wire::request_for(app);
+        let body = request
+            .to_xml()
+            .map_err(|e| NebraskaError::Serialize(e.to_string()))?;
+        trace!(
```
bfjelds, line 528: Really like how carefully secrets and protocol details are otherwise handled in this module.

Noticed that `trace!("Nebraska request to '{}':...", self.endpoint, ...)` logs the endpoint verbatim, and `error.rs` mentions the client URL may need to carry an Omaha secret. Would it be reasonable to redact the endpoint (or at least strip query/secret components) before logging it at trace level, so enabling trace diagnostics doesn't risk writing a credential into logs?

crates/trident-acl-agent/src/nebraska/README.md
frhuelsz: it's a complicated, it's technically it's a generic omaha v3 client, with a couple minor nebraska details, but i don't think  we're tied to any specific version of nebraska

crates/trident-acl-agent/src/nebraska/client.rs
bfjelds: corresponding fix applied in pr 730: https://github.com/microsoft/trident/pull/730/changes/d8b292496972616ae2bcaf389f92fc472c315f75

crates/trident-acl-agent/src/nebraska/wire.rs
frhuelsz: Pushed this decision up to the library's consumer.

crates/trident-acl-agent/src/nebraska/id.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.write_str(&self.0)
+    }
+}
+
+/// Returns whether the value is a brace-wrapped id (`{...}`), the shape Nebraska
+/// filters out.
+fn is_braced(id: &str) -> bool {
+    id.starts_with('{') && id.ends_with('}')
+}
```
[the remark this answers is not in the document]
frhuelsz: Fixed in 7ace91f9, then tightened in 4cea2395 after checking the server: Nebraska filters with `instance_id NOT LIKE '{________-____-____-____-____________}'` (`dbreads/instances.go`), and SQL `_` matches *any* character. So what it hides is the braced `8-4-4-4-12` **shape**, not a well-formed UUID — a braced non-hex id of that shape would still vanish from the UI. `is_fake_instance_shape` now matches on shape, so `{not-a-uuid}` and a braced compact UUID are both accepted, while `{zzzzzzzz-zzzz-zzzz-zzzz-zzzzzzzzzzzz}` is rejected. Covered by tests.

crates/trident-acl-agent/src/nebraska/transport.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+impl Transport for ReqwestTransport {
+    fn post_xml(&self, endpoint: &Url, body: &[u8]) -> Result<String, NebraskaError> {
+        self.client
+            .post(endpoint.as_str())
+            .header("Content-Type", "application/xml")
+            .body(body.to_vec())
+            .send()
```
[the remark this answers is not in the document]
frhuelsz: Fixed in 7ace91f9. `ReqwestTransport` now applies a per-request timeout (10s default, overridable via `ReqwestTransport::with_timeout`), so a stalled connection becomes a retryable `Transport` error instead of an unbounded hang. That makes the bounded retry actually bounded: with both defaults the worst case is under two minutes, which the `complete_after_reboot` docs now state. There is a test that points the transport at a socket which accepts but never answers, and asserts it gives up and that the error is retryable.

crates/trident-acl-agent/src/nebraska/transport.rs
frhuelsz: Good catch — same issue the automated review flagged, and you were right that it made "bounded" meaningless. Fixed in 7ace91f9: a per-request timeout (10s default, `ReqwestTransport::with_timeout` to override), so a hung DNS/TCP/TLS attempt surfaces as a retryable `Transport` error and the backoff loop gets its turn. Worst case with the defaults is now under two minutes, and the docs say so instead of hand-waving.

crates/trident-acl-agent/src/nebraska/README.md
frhuelsz: Added a "Protocol compatibility" section to the README in 62d22933, extended in 4cea2395. Rather than pinning a Nebraska release, it states this is a generic Omaha v3 client and names the specific server behaviours it depends on: events processed before the update check within one request; `error-updateInProgressOnInstance` between the first progress event and the terminal one; app/group resolved *before* events, so an unresolvable app or track drops them without an `<event>` ack; `previousversion` read from the `<event>` element; and the `<os arch>` table with its silent amd64 fallback. Those are the things that would actually break if the server changed, and unknown status strings are already tolerated by design.

crates/trident-acl-agent/src/nebraska/client.rs
frhuelsz: Fixed — and you were right that this was the silent-failure case the module is supposed to prevent. Two rounds:

62d22933 made `report_progress`/`report_failure` check the app-level status instead of just presence. Then, checking Nebraska`s source more carefully for 4cea2395, the status turned out to be the wrong signal: `buildOmahaResponse` writes app-level errors from *both* the event-resolution stage and the update-check stage. Judging by status alone meant a throttled rollout on the batched post-reboot check looked like a lost completion, and `complete_after_reboot` returned a permanent error for an update that had actually landed.

Delivery is now judged by the per-event `<event>` acknowledgements, which is the signal that genuinely distinguishes the two: when Nebraska cannot resolve the app or track it returns *before* `respApp.AddEvent()`, so there is no ack at all; once events are processed there is one per event, and any later app error belongs to the update check. `require_events_accepted` takes the expected count, and `error-updateInProgressOnInstance` is tolerated since events are sent precisely while an update is in flight.

crates/trident-acl-agent/src/nebraska/client.rs
frhuelsz: Fixed in 62d22933 — you were right, and #730 is a separate branch so the fix was still needed here.

`complete_after_reboot` now wraps the attempt so that `CheckOutcome::UpdateInProgress` becomes a retryable `NebraskaError::CompletionNotAcknowledged` inside the retry closure, rather than returning as a terminal `Ok`. So a completion that did not take is re-sent, and if it never lands the caller gets an error instead of a false success for a permanently wedged instance. `try_complete_after_reboot` keeps returning the raw outcome for callers driving their own retry, and its docs now say to loop on that outcome as well as on retryable errors. Two tests cover it: one where the second attempt succeeds, one where it never does.

crates/trident-acl-agent/src/nebraska/client.rs
frhuelsz: Fixed in 62d22933 — and it was worse than the trace line: `reqwest::Error`'s `Display` embeds the request URL, so the endpoint (and any Omaha secret in its path) was being copied into every `Transport`/`Http` error we returned, not just trace output.

Both paths are closed. Transport errors are built with `e.without_url()`, and the endpoint is logged through a `redacted()` helper that keeps only scheme and host[:port], dropping path, query, fragment and userinfo — dropped wholesale rather than scrubbed selectively, so a secret cannot leak from a component the code did not anticipate. Tests cover the query and userinfo cases, and the transport timeout test asserts the endpoint path never appears in the error text.

crates/trident-acl-agent/src/nebraska/client.rs
```diff
@@ -0,0 +1,1004 @@
...
+            "Nebraska request to '{}':\n{}",
+            redacted(&self.endpoint),
+            String::from_utf8_lossy(&body)
+        );
+        let text = self.transport.post_xml(&self.endpoint, &body)?;
+        trace!("Nebraska response:\n{text}");
+        wire::parse_response(&text).map_err(NebraskaError::Parse)
```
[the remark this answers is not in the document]
frhuelsz: Fixed in 4df3f239. The raw body is no longer logged; `send()` now traces the byte length plus a `summarize()` rendering of just the fields this client acts on — app id and status, the event acknowledgements, the update-check status, and the manifest version with a package count. No URLs, so a pre-signed `codebase` or package URL cannot reach the log. A test asserts the summary of a real offer response contains the statuses and does *not* contain the host.

crates/trident-acl-agent/src/nebraska/client.rs
```diff
@@ -0,0 +1,1142 @@
...
+/// `https://host/os.cosi` — a plausible-looking URL pointing at the wrong place.
+/// Appending the separator first keeps the codebase a directory, matching how
+/// Omaha clients concatenate the two halves.
+fn join_file(codebase: &Url, name: &str) -> Option<Url> {
+    let mut base = codebase.clone();
+    if !base.path().ends_with('/') {
+        base.set_path(&format!("{}/", base.path()));
+    }
+    base.join(name).ok()
+}
```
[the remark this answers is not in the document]
frhuelsz: Implemented this in 4df3f239 and then reverted it in 3631256a — closing as intentional.

Two reasons. It restricts something legitimate: Nebraska is a metadata server, and the `codebase` it returns is `pkg.URL`, an operator-entered base that normally points at a blob store or CDN. The artifact is *expected* to live on another host, and there is no reason a manifest cannot name it absolutely.

More importantly it defends against nothing. `name` and `codebase` arrive in the same response from the same server, so anything able to forge one can forge the other; constraining the name against the codebase is not a trust boundary. What actually protects the download is TLS to the server plus the package hash the offer already surfaces.

The trailing-slash normalization from that commit is kept, since that one is a real compatibility fix (Omaha clients concatenate codebase and name, so a codebase without the separator must not lose its last path segment) rather than a restriction.

crates/trident-acl-agent/src/nebraska/client.rs
```diff
@@ -0,0 +1,1285 @@
...
+    // opaque origins (`file:`, `data:`) never compare equal, not even to
+    // themselves, so anything that leaves the web-URL world fails closed.
+    // The path prefix then confirms the file is inside the codebase directory,
+    // and the inequality rejects a name that resolves back to the directory.
+    let contained = url.origin() == base.origin()
+        && url.path().starts_with(base.path())
+        && url.path() != base.path();
+
+    contained.then_some(url)
+}
```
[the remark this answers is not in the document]
frhuelsz: Moot — the containment check this refers to was removed in 3631256a, so there is no longer a prefix check to bypass. See the reply on the thread above for why: the package name and the codebase come from the same response, so validating one against the other is not a trust boundary, and it blocked the legitimate case of an artifact hosted on a CDN rather than on Nebraska itself. Download integrity rests on TLS plus the package hash.

crates/trident-acl-agent/src/nebraska/client.rs
```diff
@@ -0,0 +1,1241 @@
...
+/// into `https://host/os.cosi` — a plausible-looking URL pointing at the wrong
+/// place. Appending the separator first keeps the codebase a directory,
+/// matching how Omaha clients concatenate the two halves.
+fn join_file(codebase: &Url, name: &str) -> Option<Url> {
+    let mut base = codebase.clone();
+    if !base.path().ends_with('/') {
+        base.set_path(&format!("{}/", base.path()));
+    }
+    base.join(name).ok()
+}
```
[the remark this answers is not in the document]
frhuelsz: Correct that the query is dropped, but keeping the current behaviour — documented in 92c3474f rather than changed.

Checked what the protocol actually intends. Nebraska stores a package as a free-form `codebase` plus a `filename` (`types.Package.URL` / `.Filename`), validates or normalizes neither, and emits them as `<url codebase=…>` + `<package name=…>`. The reference client combines them by concatenation, not relative resolution — `update_engine/omaha_request_action.cc`:

```cpp
// Add a slash if it is not already there
if (payload_urls[i].length() > 0 && payload_urls[i].back() != 0x2F) {
  payload_urls[i] += "/";
}
payload_urls[i] += package_name;
```

So there is no relative-link concept anywhere in the protocol, and resolving is a divergence on our side. We are keeping it deliberately: it honours a `name` that is itself an absolute URL, which lets a manifest point straight at an artifact hosted off the codebase — something concatenation cannot express (it would produce `<codebase>/https://cdn/os.cosi`).

On the specific scenario: a container-level SAS on the codebase is unsupported by Omaha either way. Concatenation yields `…/pkgs/?sv=…&sig=SECRETSIGos.cosi`, which is equally broken — so this is not a regression against the reference client. The supported way to pre-sign is per file, in the name, where the query *is* preserved.

All three behaviours now have tests, and the trade-off is stated on `join_file` and in the module README so the next reader does not have to rediscover it.

bfjelds (approved):

PR #742, opened 2026-08-18
packaging: add %license LICENSE to all subpackages

frhuelsz:
## What

Add `%license LICENSE` to every `%files` section in `packaging/rpm/trident.spec` — the main `trident` package and all subpackages (`provisioning`, `service`, `install-service`, `selinux`, `static-pcrlock-files`, `acl-agent`).

## Why

Packaging guidelines require each (sub)package to ship its license file via the `%license` directive rather than relying on the license only being present in the main package. This ensures the MIT `LICENSE` is installed alongside every produced RPM (under the license dir), which is important since the subpackages can be installed independently of the main package (e.g. `selinux`, `static-pcrlock-files`).

The top-level `LICENSE` file already exists in the source tree and is present in both build paths (the `%autosetup`-unpacked tarball for the Azure Linux distro build, and the in-tree checkout for the Trident repo build), so `%license LICENSE` resolves in both.

## Validation

- `rpmspec -P` parses cleanly on both build paths (with and without `rpm_ver` defined).
- 7 `%license LICENSE` entries, one per `%files` section.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/packaging/rpm/trident.spec
+++ b/packaging/rpm/trident.spec
@@ -82,6 +82,7 @@ Trident. This package provides the Trident tool
 and its dependencies for managing the lifecycle of Azure Linux hosts.
 
 %files
+%license LICENSE
 %{_bindir}/%{name}
 %dir /etc/%{name}
 %{_unitdir}/%{name}d.service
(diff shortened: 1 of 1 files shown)


bfjelds:
Superseded by #744, which now also includes these subpackage %license additions plus the Docker build-context fix (COPY LICENSE) needed for the in-repo rpmbuild --build-in-place path.

PR #743, opened 2026-08-18
spec: package LICENSE file with %license directive

bfjelds:
trident.spec declares License: MIT but never installs the license text via %license, so the built RPM does not carry LICENSE under %{_licensedir} as required by packaging guidelines.

- Add %license LICENSE to the main package %files section.
- COPY LICENSE . into the Docker build contexts (Dockerfile.full, Dockerfile.azl, Dockerfile.full.public) that build the RPM via 
pmbuild --build-in-place, since it was not previously copied into the build context.

Verified locally: 
pmspec -q parses the updated spec correctly, and a docker build using Dockerfile.azl progresses through %install/packaging (finding and packaging LICENSE) without error; the only failure seen was in %check due to a placeholder test binary unrelated to this change.

### DIFF
--- a/packaging/docker/Dockerfile.azl
+++ b/packaging/docker/Dockerfile.azl
@@ -19,6 +19,7 @@ RUN tdnf install -y \
 WORKDIR /work
 
 COPY packaging/rpm/trident.spec .
+COPY LICENSE .
 COPY packaging ./packaging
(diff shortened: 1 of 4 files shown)


bfjelds:
Superseded by a PR from user/bfjelds/fix-spec-license-directive (pushed directly to this repo).

PR #745, opened 2026-08-18
packaging: ship LICENSE + third-party OSS NOTICE in RPMs (with cargo-about drift check)

frhuelsz:
## What

Combines the two closely-related license/packaging efforts into one change (subsumes #744, thanks @bfjelds):

1. Ship the first-party `LICENSE` in **every** RPM subpackage via `%license`.
2. Generate a top-level `NOTICE` (third-party OSS attribution) with `cargo-about`, ship it in the two packages that bundle third-party crates, and guard it against drift in CI.

## Why

trident.spec declares `License: MIT` but never installed the license text, so built RPMs did not carry `LICENSE` under `%{_licensedir}`. Separately, because Rust links dependencies statically, the shipped binaries bundle ~150 third-party crates (MIT / Apache-2.0 / BSD-3-Clause / Unicode-3.0 / Zlib per `deny.toml`), all of which require their attribution to travel with the binary. `cargo-deny` gates the *allowed* license set but produces no attribution — this fills that gap.

### LICENSE in the RPMs (from #744)
- `trident.spec`: `%license LICENSE` on every `%files` section (main + all subpackages).
- `COPY LICENSE …` into the Docker build contexts (`Dockerfile.azl`, `Dockerfile.full`, `Dockerfile.full.public`), which build via `rpmbuild --build-in-place` and did not previously stage it. Also picks up #744's stale-path fixes in `Dockerfile.full.public` (`trident.spec` path, `.cargo/config.toml`).

### Third-party NOTICE (cargo-about)
- `about.toml` / `about.hbs`: config + plain-text template. `accepted` synced with `deny.toml`; `targets` pinned to the two supported linux triples so output is host-independent; build-/dev-dependencies and first-party (`publish = false`) crates excluded.
- `NOTICE`: generated, checked in. One section per unique license text with the crates that use it (correct granularity for MIT/BSD per-crate copyright and Apache NOTICE preservation).
- `trident-proto`: add `publish = false` for consistency with the other first-party crates, so it's excluded from the notice.
- `%license LICENSE NOTICE` on the two packages that bundle third-party code (main `trident` and `trident-acl-agent`); the unit/policy/data subpackages get `LICENSE` only and receive `NOTICE` transitively via `Requires: trident`.

### Drift check (mirrors the API-schema check)
- `make update-notice` — regenerate the checked-in `NOTICE` locally (locked).
- `make validate-notice` — regenerate to a temp file and `diff` against the checked-in `NOTICE`.
- The `Check Licenses` CI job (runs on PR next to `cargo deny check licenses`) installs a pinned `cargo-about@0.9.1` and runs `make validate-notice`, failing on drift.

## Validation

- `make update-notice` is idempotent; `make validate-notice` passes clean and fails on a stale `NOTICE`. No first-party crates / no HTML-escape artifacts in the output. Spec parses on both build paths.
- **End-to-end container build** (`Dockerfile.azl`, run on the merged tree): the produced RPMs install
  `/usr/share/licenses/trident/{LICENSE,NOTICE}` and `/usr/share/licenses/trident-acl-agent/{LICENSE,NOTICE}` **byte-identical to source**, with `LICENSE`-only for the remaining subpackages (verified via `rpm -qLp`).

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


about.toml
```diff
@@ -0,0 +1,40 @@
...
+# The accepted license set is kept in sync with `deny.toml`'s `[licenses] allow`
+# list; a new license appearing in the tree must be vetted and added to both.
+
+# SPDX identifiers accepted for third-party crates. Keep in sync with deny.toml.
+accepted = [
+    "MIT",
+    "Apache-2.0",
+    "BSD-3-Clause",
+    "Unicode-3.0",
+    # `const_format` / `const_format_proc_macros` are Zlib (see deny.toml exceptions).
```
fintelia:
```suggestion
```

about.toml
```diff
@@ -0,0 +1,40 @@
...
+# Pin the target triples so the generated dependency graph — and therefore the
+# NOTICE — is identical regardless of the host that runs the generation.
+targets = [
+    "x86_64-unknown-linux-gnu",
+    "aarch64-unknown-linux-gnu",
+]
+
+# Only crates whose code is actually linked into a shipped binary need
+# attribution: build-time-only and test-only dependencies are not bundled.
+ignore-build-dependencies = true
```
fintelia: Build dependencies can sometimes land in the final binary, so maybe don't want to ignore them.

about.toml
```diff
@@ -0,0 +1,40 @@
...
+# linked into the Trident binaries.
+#
+# `NOTICE` is checked in and regenerated in CI to detect drift (see the
+# `update-notice` / `validate-notice` Makefile targets and the "Check Licenses"
+# pipeline job). To regenerate locally after a dependency change, run:
+#
+#   make update-notice
+#
+# The accepted license set is kept in sync with `deny.toml`'s `[licenses] allow`
+# list; a new license appearing in the tree must be vetted and added to both.
```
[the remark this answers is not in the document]
frhuelsz: Good catch — reworded to clarify that the accepted set mirrors deny.toml's `allow` list **plus** its per-crate `exceptions` (e.g. Zlib for const_format), rather than just the allow list. Also dropped the now-redundant inline const_format comment. (b93ec11f)

about.toml
frhuelsz: Done — removed the inline `const_format` … Zlib comment; the header comment now covers why Zlib is accepted. (b93ec11f)

about.toml
frhuelsz: Agreed — flipped to `ignore-build-dependencies = false` so build scripts / proc-macros that can emit code into the final binary are attributed. Regenerated NOTICE picks up the build-dep crates (cc, prost-build, tonic-build, …). (b93ec11f)

bfjelds (dismissed):

bfjelds (approved):

PR #747, opened 2026-08-18
telemetry: emit image_read_complete tracing metric

bfjelds:
ReadMonitor collects great metrics for image downloads. Add a
completion metric emitted through the existing tracing metrics
pipeline, fired via Drop. Tag events with a label for context.

### DIFF
--- a/crates/trident/src/engine/storage/image.rs
+++ b/crates/trident/src/engine/storage/image.rs
@@ -206,9 +206,12 @@ pub(super) fn deploy_images(ctx: &EngineContext) -> Result<(), TridentError> {
 }
 
 /// Mount points allow-listed for use as `ReadMonitor` metric labels. Any
-/// mount point not on this list (including partitions with no associated
-/// mount point, e.g. verity hash devices or raw COSI partitions) is reported
-/// as "other". This keeps metric cardinality bounded and avoids leaking
+/// mount point not on this list is reported as "other". Partitions with no
+/// associated mount point at all (e.g. raw COSI partitions) are also always
+/// reported as "other". Verity hash devices are the exception: they don't
+/// have their own mount point, but are tagged with the mount point label of
+/// the data device they protect, since they're deployed as part of the same
+/// filesystem. This keeps metric cardinality bounded and avoids leaking
 /// user-defined block device ids or arbitrary mount paths into the metrics
 /// pipeline.
 fn metric_label_for_mount_point(mount_point: &Path, esp_mount_path: &Path) -> &'static str {
--- a/crates/trident/src/engine/storage/image.rs
+++ b/crates/trident/src/engine/storage/image.rs
@@ -175,12 +175,12 @@ pub(super) fn deploy_images(ctx: &EngineContext) -> Result<(), TridentError> {
 
         let monitored_reader = ReadMonitor::new(
             reader,
-            metric_label,
+            id.clone(),
             image_file.compressed_size,
             threshold_reporting,
             reporting_interval,
         )
-        .make_reporting("image_read_complete");
+        .make_reporting("image_read_complete", metric_label);
 
         if let Err(e) = deploy_os_image_file(ctx, &id, image_file, resize, monitored_reader) {
             return ControlFlow::Break(Err(e).structured(ServicingError::DeployImages));
--- a/crates/trident/src/engine/storage/image.rs
+++ b/crates/trident/src/engine/storage/image.rs
@@ -89,6 +89,8 @@ pub(super) fn deploy_images(ctx: &EngineContext) -> Result<(), TridentError> {
             .structured(InternalError::Internal("No image found for mount point"))
             .message(format!("Mount point '{}' should have image", mpp.display()))?;
 
+        let metric_label = metric_label_for_mount_point(&mpp, ctx.esp_mount_path.as_path());
+
         // Check if this ID is a verity device, if so, we must explore the graph
         // to obtain the underlying devices.
         if let Some(verity_device) = ctx.spec.storage.verity_device(&id) {
@@ -104,6 +106,7 @@ pub(super) fn deploy_images(ctx: &EngineContext) -> Result<(), TridentError> {
                 image.image_file.path.clone(),
                 (
                     verity_device.data_device_id.clone(),
+                    metric_label,
                     &image.image_file,
                     FileSystemResize::NoResize,
                 ),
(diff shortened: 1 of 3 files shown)


crates/trident/src/io_utils/read_monitor.rs
```diff
@@ -10,11 +10,17 @@ const RING_BUFFER_SIZE: usize = 10;
...
+/// configurable minimum cadence. When the wrapper is dropped (i.e. the read
+/// is finished, successfully or not), it emits a single `tracing` metric
+/// event (picked up by the metrics pipeline) with the overall size and
+/// average rate for the whole read.
 pub struct ReadMonitor<R> {
     inner: R,
-    /// Expected size of the complete file being read (for log context).
+    /// Identifies what is being read, for metric/log context (e.g. a
+    /// partition id, or "esp").
+    label: String,
```
bfjelds, line 26: should label be included?

crates/trident/src/io_utils/read_monitor.rs
```diff
@@ -28,19 +34,32 @@ pub struct ReadMonitor<R> {
...
-    /// * `threshold_mbps` — speed in Mbps below which debug messages are
-    ///   emitted.
-    /// * `report_cadence` — minimum interval between consecutive log messages.
-    pub fn new(inner: R, size: u64, threshold_mbps: f64, report_cadence: Duration) -> Self {
+    /// * `label` — identifies what is being read (e.g. partition id), used to
+    ///   tag log messages and metric events.
+    /// * `threshold_mbps` — speed in Mbps below which debug messages/metrics
+    ///   are emitted.
+    /// * `report_cadence` — minimum interval between consecutive log
+    ///   messages/metrics.
```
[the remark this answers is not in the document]
bfjelds: Great catch -- fixed the docs for `threshold_mbps` and `report_cadence` to clarify they only gate the slow-download debug log, not the summary metric (which is always emitted once at drop time regardless of speed). See 313741ac.

crates/trident/src/io_utils/read_monitor.rs
frhuelsz: To keep this a generic monitor, I'd prefer to keep this functionality non-default and configurable, we mostly just use this to download images, but it's a generic IO utility and any future use could accidentally pollute our metrics.
I'd make the label and the metric production optional, or enable them via a monitor wrapper.
eg
```rust
// optional
let monitor = ReadMonitor::new(...)
    .with_label(...)
    .with_metric(<metric_name>);
// or 
let monitor = ReadMonitor::new(...)
    .with_metric(<label>, <metric_name>);

// wrapper
let monitor = ReadMonitor::new(...)
    .make_reporting(<label>,<metric_name>);
// this would return an instance of a new ReportingReadMonitor type, that wraps the ReadMonitor
```

Another approach would be to not change the monitor itself, but rather the callers to not lose the monitor object and produce the metric from the data it collected.

```rust
// I think this is ok?
let mut monitor = ReadMonitor::new(...);
read_operation(&mut monitor);

// Now do metric
tracing::info!(
    <some_field> = monitor.<some_field>
    ...
);
```

frhuelsz: As for sending the block device id... I'd rather send the mount path, but perhaps limit it to an allow-list of common things? All solutions seem not great.

bfjelds: i like `make_reporting` and will start with allow-list: root|usr|esp|boot else other

bfjelds: label is still passed into ReadMonitor for logging, not tracing.

allow-list sanitized label is passed to ReportingReadMonitor for tracing.

crates/trident/src/engine/storage/image.rs
```diff
@@ -112,6 +115,7 @@ pub(super) fn deploy_images(ctx: &EngineContext) -> Result<(), TridentError> {
                 image_file_verity.hash_image_file.path.clone(),
                 (
                     verity_device.hash_device_id.clone(),
+                    metric_label,
                     &image_file_verity.hash_image_file,
                     FileSystemResize::NoResize,
                 ),
```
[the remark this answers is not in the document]
bfjelds: Great comment. Fixed -- the doc comment incorrectly lumped verity hash devices in with the truly-unmounted partitions. Updated it to clarify that hash devices inherit the mount point label of the data device they protect.

crates/trident/src/io_utils/read_monitor.rs
```diff
@@ -138,6 +204,52 @@ impl<R: Read> Read for ReadMonitor<R> {
...
+
+        tracing::info!(
+            metric_name = self.metric_name,
+            label = self.label,
+            bytes_read = self.monitor.total_bytes,
+            total_bytes = self.monitor.size,
+            duration_secs = duration_secs,
+            avg_mbps = avg_mbps,
+            complete = complete,
+        );
```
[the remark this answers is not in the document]
bfjelds: Great comment. This is a false alarm -- `String` already implements `tracing::field::Value`, so `self.metric_name` / `self.label` are recorded by reference (no move out of `&mut self`). This has compiled cleanly across every build/test/clippy -D warnings run I did while developing this change, including after this exact code was added.

crates/trident/src/io_utils/read_monitor.rs
```diff
@@ -10,11 +10,18 @@ const RING_BUFFER_SIZE: usize = 10;
...
+///
+/// `ReadMonitor` itself is a generic IO utility and never emits metrics. To
+/// additionally get a single summary `tracing` metric event when the read
+/// finishes (successfully or not), wrap it with [`ReadMonitor::make_reporting`].
 pub struct ReadMonitor<R> {
     inner: R,
-    /// Expected size of the complete file being read (for log context).
+    /// Identifies what is being read, for metric/log context (e.g. a
+    /// partition id, or "esp").
+    label: String,
```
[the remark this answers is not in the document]
bfjelds: Great comment. Fixed at all flagged spots -- the `label` field doc and the `new()` constructor param doc were both stale from before `ReadMonitor` stopped emitting metrics directly. They now correctly say `label` only tags the local slow-download debug log, while the metric label (if any) comes separately from `make_reporting`. The `size`/`total_bytes` `log/metric context` doc comments were left as-is since those fields genuinely are used in both the debug log and the metric.

frhuelsz (approved):

PR #754, opened 2026-08-27
engineering: make Azure Container Linux images installable (inline dm-verity, ESP staging)

frhuelsz:
# 🔍 Description

**engineering + bug.** Makes Azure Container Linux (ACL) images installable.

Three defects, all found installing the same image:

1. **Inline dm-verity is not supported when deriving a Host Configuration from COSI**, so `trident grpc-client stream-disk` cannot install ACL. Inline verity stores the hash tree inside the same partition as the data, at a byte offset, rather than in a dedicated hash partition.
2. **The verity-addon guard keys off a shared directory**, so `trident install` rejects every currently published ACL image.
3. **fstab is written for UKI usr-verity images**, which both fails on ACL and would be wrong if it succeeded.

> A fourth defect from the same effort — file-based ESP deployment assuming `<newroot>/var/tmp` exists — was split out and merged separately as #790, since it stands alone and affects any image lacking that directory. This PR is rebased on top of it.

They are separate faults, uncovered one behind the other as each was fixed. `stream-disk` only hits (1); `install` hits all three. None alone makes ACL installable.

# 🤔 Rationale

Streaming a real ACL image failed during derivation:

```
Derived Host Configuration is invalid: Referrer 'verity-1' of kind
'verity-device' references target 'partition-2' more than once
```

Derivation resolved the verity hash partition purely by image path. With inline verity the verity entry points at the *same* image as the filesystem, so it emitted a `VerityDevice` whose data and hash devices were the same partition, which the storage graph rejected. Trident also had nowhere to record where the hash tree began, and never passed an offset to `veritysetup open`.

This was not a regression — inline verity was never supported. `git log --all --grep=inline` is empty and nothing referenced `hashOffset`. #673 made ACL work for **A/B update**, but that path never derives a Host Configuration; it reads the root hash from `usrhash=` on the UKI cmdline and mounts `/usr` directly, so it never had to model the layout.

### Inline dm-verity (1)

- **Read `hashOffset` from COSI verity metadata.** Image Customizer already emits this field for inline layouts — it is declared in its `cosiapi` package and was added upstream by [azure-linux-image-tools#702](https://github.com/microsoft/azure-linux-image-tools/pull/702) ("Add support for inline verity"), gated so it appears only when inline, and covered by an IC test. The encoding is intentional and stable on the producer side, confirmed against IC's source: `HashOffset` is set **only** when the hash tree is inline (`cosicommon.go:199-209`), and the separate-hash layout instead yields a *distinct* image path with the field omitted. So the two shapes are cleanly discriminable, and `verity.image.path == image.path` together with a present `hashOffset` is precisely how IC expresses inline verity. It has behaved this way in every release since v1.4.0 (April 2026) with no COSI version bump, so **upgrading IC does not avoid this** — a consumer that wants to install inline-verity images has to read the field.

This change consumes what the producer already declares, and brings Trident's COSI reference up to date with it: `hashOffset` is now documented in `Composable-OS-Image.md`, the v1.2 JSON Schema, and the changelog. The addition is descriptive and backward compatible — the field is optional and absent for the ordinary separate-hash-partition layout, so every existing COSI stays valid.
- **Carry the offset on the OS image, not the Host Configuration.** Like the root hash, the offset is a property of the image, so it is read from image metadata at servicing time. It does **not** appear in the Host Configuration or Host Status.
- **Allow a verity device to name the same partition as both data and hash device.** `hashDeviceId` stays required, so the Host Configuration schema is **completely unchanged** — `host-config-schema.json` and the generated `VerityDevice.md` are byte-identical to `main`. The graph models the inline case as the single device it is rather than as a second reference to the same node, which leaves the duplicate-target and referrer-sharing invariants untouched. Verity referrer cardinality becomes `1..=2`, and the hash partition-type cross-check is skipped when there is no distinct hash partition.
- **Pass `--hash-offset` to `veritysetup open`,** so an inline-verity device is actually opened and verified during servicing rather than skipped.

Two documentation sets are treated differently on purpose. The **COSI reference** describes the format Trident consumes, so it documents `hashOffset` as a factual property of images IC already produces. The **Host Configuration reference** describes what users may configure, and inline dm-verity is **not an officially supported configuration**, so it is deliberately not documented there — `host-config-schema.json` and the generated `VerityDevice.md` are byte-identical to `main`. The only generated-doc change is the verity-device row of the referrer cardinality table in `Storage-Rules.md` (`2..2` → `1..2`), which states a reference count without mentioning inline verity.

Derivation fails with a clear error naming `hashOffset` if an image uses inline verity but does not say where the hash tree starts, instead of producing a device that cannot be opened.

### Verity addon guard (2)

`activate_verity_addon_for_target_volume` decided whether an image uses PARTUUID-based verity addons by testing whether `acl/uki-addons/` exists, then hard-errored if the selected slot's template was missing. That directory is shared — its own doc comment says it holds "per-slot verity addons (`verity-a`/`verity-b`) and the first-boot addon" — and published ACL images ship first-boot, fips and kdump there while shipping **no** verity templates:

```
$ mdir -i esp.img -/ ::/acl/uki-addons     # ESP straight out of the ACL COSI
 FIRSTB~1 EFI     81784   firstboot.addon.efi
 FIPSAD~1 EFI     81784   fips.addon.efi
 KDUMPA~1 EFI     81784   kdump.addon.efi
```

So the guard passed and the function demanded a template the image never had, failing installation of every published ACL image. Per-slot verity addons are ACL base work that has not landed yet.

The decision now keys on whether the image's verity is **inline**. A per-slot addon exists to carry its slot's verity data/hash PARTUUID pair; inline verity has no such pair, since the hash tree lives in the data partition and both refer to one device, so activation has nothing to select and is skipped.

Gating on inline-ness rather than on "no templates present" matters: the latter infers intent from a missing file, so an image that *should* ship per-slot addons but shipped none by mistake would be indistinguishable from one legitimately exempt, and would be activated silently instead of rejected. Inline-ness is a property that can be established positively.

With a separate hash partition the check therefore stays strict — including when no templates are shipped at all, which is a packaging fault rather than an exemption. That is the direction ACL is heading, and where the error should keep firing. The pre-existing safety test (ships `verity-a`, installs to slot B → error) is unchanged and still passes.

Inline-ness comes from `VerityDevice::is_inline()`, the same predicate defect (1) keys on, so the two features stay consistent by construction.

### fstab on UKI usr-verity images (3)

Trident already skips storage configuration for UKI images using **root**-verity, on the reasoning that such an image takes its mount topology from the signed kernel command line rather than from fstab. ACL is that same case with the verity on `/usr`, but the guard tested only root-verity, so it did not apply and `generate_fstab` ran — failing, because ACL's ROOT image ships no `/etc`.

Creating the directory would have been the wrong fix. ACL assembles `/etc` as an overlay — factory `/usr/share/distro/etc` as lower, the ROOT partition's `/etc` as upper — so anything written there wins over the image's own. Confirmed against the shipping image:

- the factory `/usr/share/distro/etc/fstab` is deliberately **empty** (0 bytes);
- every mount ACL needs already comes from the command line (`/usr`, `/`) or a shipped unit (`/boot`, `/oem`);
- `systemd-fstab-generator` is present, and emits into `/run/systemd/generator`, which takes precedence over `/usr/lib/systemd/system`.

A generated fstab would therefore have **shadowed the units the image ships**, not merely duplicated them. The absence of `/etc` is likewise expected rather than a fault: ACL's initrd creates it immediately before mounting the overlay, in dracut module `99setup-root`.

Only fstab is skipped, not the whole step as in the root-verity case: a usr-verity image still has a writable root, so encryption and RAID configuration remain applicable.

# 📝 Checks

- [x] Check [dev-docs/manual-validation.md](/dev-docs/manual-validation.md)

Both defects were found and re-validated on real ACL images. Manually validated end to end on a real ACL image (`acl-base-3.0.20260706`), installed via `stream-disk` from an Azure Linux installer ISO built with RPMs from this branch, in QEMU/KVM:

- derivation succeeds; storage graph builds; full `HostConfiguration::validate()` passes
- all five partitions streamed; `Servicing completed successfully`; reboot
- verity activated during servicing and reported `verified`
- **installed disk boots with no install media attached:** UEFI → GRUB → ACL kernel `6.6.143.1`, with systemd activating dm-verity on `/usr` from the UKI cmdline (`sysusr-usr.mount`, `/dev/mapper/usr`, zero mount failures), 38 targets reached

First boot then stops at `emergency.target` because Ignition cannot reach Azure IMDS (`169.254.169.254`) — expected off-Azure for an image carrying `flatcar.oem.id=azure`, and downstream of everything this PR touches (verity was already mounted and verified before Ignition ran).

Before writing any code, the underlying `veritysetup` invocation was confirmed against the real disk image:

```
veritysetup open /dev/loopXp2 <name> /dev/loopXp2 270ed371... --hash-offset=1065345024
→ status: verified
```

Automated: 22 test suites pass workspace-wide, `cargo fmt --check` clean, `cargo clippy` introduces no new warnings. Adds `test_derive_host_configuration_inner_inline_verity` (pinning ACL's real `usr` partition type GUID, which exercises partition-type rules a generic Linux GUID would mask) and `..._inline_verity_missing_offset`.

# 📌 Follow-ups

TODO:

- The v1.2 schema requires `osPackages` to be an array, but real ACL COSI files set it to `null` and Trident already tolerates that (#707). Validating an ACL image's metadata against the checked-in schema fails on `osPackages` alone — a pre-existing divergence, unrelated to this PR and left alone here.

# 🗒️ Notes

Opened as a **draft**: functionally validated end to end, but wanted review on the graph modelling choice before marking ready.

The main design question is how inline verity is represented. Two alternatives were built and discarded:

1. **Drop the verity device entirely**, mounting `/usr` read-only straight off the partition and leaving verity to systemd at boot. Smallest change and it validates, but Trident never verifies the image it just wrote. It also has a trap: the non-verity path drops the `ro` mount option, and `resize` is skipped only when `is_read_only() || !fs_type.is_ext()`. ACL survives on the `btrfs` clause alone; an ext4 inline-verity image would be resized and have its appended hash tree silently destroyed.
2. **`hash_device_id: Option<BlockDeviceId>`.** Cleaner conceptually, but it is a Host Status wire-format change that rippled through ~30 files.

The approach here keeps the Host Configuration and Host Status untouched while still verifying at install time.

### DIFF
--- a/crates/osutils/src/fatlabel.rs
+++ b/crates/osutils/src/fatlabel.rs
@@ -1,0 +1,97 @@
+//! Thin wrapper around `fatlabel`, which reads and writes the volume label of a
+//! FAT filesystem.
+//!
+//! `mkfs.vfat` can only set a label at creation time, via `-n`, so this is the
+//! way to label a FAT filesystem that already exists.
+
+use std::path::Path;
+
+use anyhow::{ensure, Error};
+
+use crate::dependencies::Dependency;
+
+/// Maximum length of a FAT volume label, in characters, as enforced by
+/// `fatlabel` itself.
+pub const MAX_LABEL_LENGTH: usize = 11;
+
+/// Returns an error if `label` will not fit in a FAT volume label.
+///
+/// Separate from [`set_label`] so the rule can be exercised without invoking
+/// `fatlabel`.
+fn validate_label(label: &str) -> Result<(), Error> {
+    ensure!(
+        label.chars().count() <= MAX_LABEL_LENGTH,
+        "FAT volume label '{label}' is longer than the {MAX_LABEL_LENGTH} characters a FAT \
+        filesystem can hold"
+    );
+    Ok(())
+}
+
+/// Sets the volume label of the FAT filesystem at `device_path`.
+///
+/// The device may be mounted; the label is written to the boot sector and
+/// survives the filesystem being unmounted.
+pub fn set_label(device_path: impl AsRef<Path>, label: impl AsRef<str>) -> Result<(), Error> {
+    let device_path = device_path.as_ref();
+    let label = label.as_ref();
+
+    validate_label(label)?;
+
+    Dependency::Fatlabel
+        .cmd()
+        .arg(device_path)
+        .arg(label)
+        .run_and_check()
+        .map_err(Error::from)
+}
+
+#[cfg(feature = "functional-test")]
+#[cfg_attr(not(test), allow(unused_imports, dead_code))]
+mod functional_test {
+    use super::*;
+
+    use pytest_gen::functional_test;
+
+    use crate::{blkid, filesystems::MkfsFileSystemType, mkfs};
+
+    #[functional_test(feature = "helpers")]
+    fn test_set_label() {
+        let device = Path::new("/dev/sda1");
+        mkfs::run(device, MkfsFileSystemType::Vfat).unwrap();
+
+        set_label(device, "TESTLABEL").unwrap();
+        assert_eq!(blkid::get_filesystem_label(device).unwrap(), "TESTLABEL");
+    }
+
+    #[functional_test(feature = "helpers", negative = true)]
+    fn test_set_label_too_long() {
+        set_label(Path::new("/dev/sda1"), "THIS-LABEL-IS-TOO-LONG").unwrap_err();
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_validate_label_rejects_overlong_label() {
+        let err = validate_label("THIS-LABEL-IS-TOO-LONG").unwrap_err();
+        assert!(err.to_string().contains("longer than"), "got: {err}");
+    }
+
+    #[test]
+    fn test_validate_label_accepts_the_maximum_length() {
+        // The conventional ESP label is exactly at the limit, so the boundary
+        // is worth pinning.
+        let label = "EFI-SYSTEM!";
+        assert_eq!(label.chars().count(), MAX_LABEL_LENGTH);
+        validate_label(label).unwrap();
+    }
+
+    #[test]
+    fn test_validate_label_counts_characters_not_bytes() {
+        // 11 multi-byte characters fit; a byte-length check would reject them.
+        validate_label("ÄÄÄÄÄÄÄÄÄÄÄ").unwrap();
+        validate_label("ÄÄÄÄÄÄÄÄÄÄÄÄ").unwrap_err();
+    }
+}
--- a/crates/osutils/src/blkid.rs
+++ b/crates/osutils/src/blkid.rs
@@ -36,6 +36,15 @@ pub fn get_partition_label(device_path: impl AsRef<Path>) -> Result<String, Erro
     run(device_path, "PARTLABEL")
 }
 
+/// Returns the filesystem label of the filesystem at `device_path`, which may
+/// be a block device or a filesystem image file.
+///
+/// This is the label `/dev/disk/by-label/` is built from, and is distinct from
+/// the GPT partition name returned by [`get_partition_label`].
+pub fn get_filesystem_label(device_path: impl AsRef<Path>) -> Result<String, Error> {
+    run(device_path, "LABEL")
+}
+
 #[cfg(feature = "functional-test")]
 #[cfg_attr(not(test), allow(unused_imports, dead_code))]
 mod functional_test {
--- a/crates/osutils/src/dependencies.rs
+++ b/crates/osutils/src/dependencies.rs
@@ -99,6 +99,7 @@ pub enum Dependency {
     Efivar,
     Efibootmgr,
     Eject,
+    Fatlabel,
     Findmnt,
     #[strum(serialize = "grub2-mkconfig")]
     Grub2Mkconfig,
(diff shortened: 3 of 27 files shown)


crates/osutils/src/fatlabel.rs
```diff
@@ -0,0 +1,92 @@
...
+        set_label(Path::new("/dev/sda1"), "THIS-LABEL-IS-TOO-LONG").unwrap_err();
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_set_label_rejects_overlong_label() {
```
frhuelsz: these two must be funtional tests as they depend on calling a system binary, they leave the sandbox

frhuelsz: Good catch, and the second one was the real offender: `"EFI-SYSTEM!"` passes the length check, so `set_label` went on to actually invoke `fatlabel` on `/dev/null`. That leaves the sandbox exactly as you say.

Rather than move both across, I pulled the length rule out into a pure `validate_label`, so the unit tests exercise the rule and cannot reach a binary however the code changes later — the first test was only sandbox-safe by accident of check ordering, which would have been easy to break.

Unit tests now cover the rule only (reject too long, accept exactly 11, and count characters rather than bytes — a byte check would wrongly reject 11 multi-byte characters). The real labelling stays in the existing functional tests, `test_set_label` and `test_set_label_too_long`, which already cover both the positive and negative paths against a real device.

PR #761, opened 2026-08-31
acl: surface Trident error (kind/subkind/location) on the update-stated annotation

bfjelds:
Adds an optional error object to the ACL A/B update-status annotation schema, populated from TridentClientError::Remote when a stage, finalize, rollback, or commit operation fails against tridentd. The object carries kind, subkind, and an optional source location so downstream consumers can distinguish and correlate Trident-originated failures without parsing the free-form message field.

Validated: https://dev.azure.com/mariner-org/ECF/_build/results?buildId=1195298&view=logs&j=09bab65c-0c5c-54aa-a756-16c7a376f940&t=09bab65c-0c5c-54aa-a756-16c7a376f940

### DIFF
--- a/crates/trident-acl-agent/src/annotations/protocol.rs
+++ b/crates/trident-acl-agent/src/annotations/protocol.rs
@@ -970,7 +970,8 @@ mod tests {
                 "object" => value.is_object(),
                 "array" => value.is_array(),
                 "boolean" => value.is_boolean(),
-                "number" | "integer" => value.is_number(),
+                "number" => value.is_number(),
+                "integer" => value.is_i64() || value.is_u64(),
                 other => panic!(
                     "test schema validator does not support type {other:?} - extend schema_validate_property"
                 ),
@@ -1422,6 +1423,11 @@ mod tests {
         too_large["tridentError"]["location"]["line"] = serde_json::json!(4294967296u64);
         schema_validate(&schema, &too_large)
             .expect_err("schema must reject a line number beyond u32::MAX");
+
+        let mut fractional = base.clone();
+        fractional["tridentError"]["location"]["line"] = serde_json::json!(1.5);
+        schema_validate(&schema, &fractional)
+            .expect_err("schema must reject a fractional line number");
 
         schema_validate(&schema, &base)
             .expect("baseline instance with an in-range line must conform to the schema");
--- a/crates/trident-acl-agent/src/annotations/protocol.rs
+++ b/crates/trident-acl-agent/src/annotations/protocol.rs
@@ -19,7 +19,16 @@ use serde::{Deserialize, Serialize};
 use url::Url;
 use uuid::Uuid;
 
-use crate::core::{config::DEFAULT_ANNOTATION_PREFIX, error::AgentError};
+use crate::core::{
+    config::DEFAULT_ANNOTATION_PREFIX, error::AgentError, trident::TridentClientError,
+};
+
+/// Sentinel `kind` used when Trident reported a remote failure without any
+/// structured error at all (see `client.rs`'s `UNKNOWN_REMOTE_ERROR_SUBKIND`
+/// for the matching `subkind` sentinel) - keeps `TridentErrorInfo::kind`
+/// non-optional instead of adding a third `Option` layer for a case that
+/// should already be rare/a Trident-side contract violation.
+const UNKNOWN_ERROR_KIND: &str = "unknown";
 
 /// Suffix (appended to the configured annotation prefix) for the request
 /// annotation, e.g. `acl.microsoft.com/update-request`.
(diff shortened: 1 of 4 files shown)


crates/trident-acl-agent/src/annotations/protocol.rs
```diff
@@ -136,6 +145,8 @@ pub struct UpdateStatus {
     pub code: StatusCode,
     pub message: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub error: Option<TridentErrorInfo>,
```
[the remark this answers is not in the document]
bfjelds: there are no consumers of any previous release of this.  there is no backwards compat concern currently.

docs/Explanation/Trident-ACL-Agent.md
```diff
@@ -179,6 +179,13 @@ A status annotation (`<prefix>/update-status` or
   match on (it may include error detail that varies run to run).
 - `fromVersion`/`toVersion` are the versions the operation moved between
   (`toVersion` is absent for `rollback`, whose target is implicit).
+- `tridentError` is present only when the failure originated from a
+  structured Trident remote error (a subset of `OperationFailed` failures;
+  most `OperationFailed` statuses - and every other code - carry no
+  `tridentError`). It has `kind`/`subkind` (Trident's own error
```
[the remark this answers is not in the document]
bfjelds: Correction to my earlier reply: a7457db0 documented the new field but got the scope wrong. Fixed properly in 5d9013be - tridentError is not limited to OperationFailed (the post-reboot commit call also attaches it on TargetBootFailed when that RPC itself returns a Trident remote error), and the doc now also covers the unknown/unknown sentinel used when Trident reports failure without a structured error payload.

crates/trident-acl-agent/src/annotations/protocol.rs
```diff
@@ -757,6 +817,26 @@ mod tests {
     "operation":     { "type": "string", "enum": ["stage", "finalize", "rollback", "commit"] },
     "code":          { "type": "string", "enum": ["InProgress", "Success", "AlreadyAtTarget", "NotStaged", "OperationFailed", "TargetBootFailed", "AgentInternalError", "InvalidRequest"] },
     "message":       { "type": "string", "maxLength": 2048 },
+    "error": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["kind", "subkind"],
+      "description": "Structured Trident error, present only when the failure originated from a Trident remote error - a subset of failure codes carry this; most do not.",
```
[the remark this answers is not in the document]
bfjelds: Follow-up: added a doc pass in a7457db0 (docs/Explanation/Trident-ACL-Agent.md) covering the new tridentError field, its shape, and a failure example.

crates/trident-acl-agent/src/annotations/protocol.rs
```diff
@@ -950,6 +1037,22 @@ mod tests {
                 ));
             }
         }
+        if let Some(minimum) = prop_schema.get("minimum").and_then(Value::as_f64) {
```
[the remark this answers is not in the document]
bfjelds: Great comment, fixed in 950b67ea: the integer type check now requires an exact i64/u64 representation (value.is_i64() || value.is_u64()) instead of accepting any JSON number, so a fractional line like 1.5 is rejected. Added a fractional case to the regression test.

crates/trident-acl-agent/src/annotations/protocol.rs
```diff
@@ -757,6 +817,26 @@ mod tests {
...
+        "kind":    { "type": "string", "description": "TridentError kind, e.g. SERVICING_ERROR, HEALTH_CHECKS_ERROR." },
+        "subkind": { "type": "string", "description": "Finer-grained identifier within kind, e.g. ab-update-reboot-check." },
+        "location": {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["path", "line"],
+          "description": "Location in Trident's source where the error originated.",
+          "properties": {
+            "path": { "type": "string" },
+            "line": { "type": "integer" }
```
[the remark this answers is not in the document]
bfjelds: Great comment, fixed in 37dcd210: constrained line to 0..=4294967295 in the formal schema and extended schema_validate_property with minimum/maximum support. Added a regression test (design_doc_status_schema_rejects_out_of_range_line) covering negative and > u32::MAX values.

crates/trident-acl-agent/src/annotations/protocol.rs
frhuelsz, line 806: i hadn't noticed this is in here, not blocking but def in a follow up we should get this out into its own file

frhuelsz, line 866: is this test code?
again, not blocking, but this should use _the_ [jsonchema](https://docs.rs/jsonschema/latest/jsonschema/) crate

bfjelds: yes, is test code

bfjelds: (test code starts line ~365)

bfjelds: happy to follow up with a better way to test what we produce against the spec's schema!

bfjelds: agreed!  happy to follow up with a better way to test what we produce against the spec's schema.

frhuelsz (approved):

PR #773, opened 2026-09-03
datastore: add generic key-value storage

bfjelds:
Add a generic keyvalue table to the SQLite datastore so arbitrary structured data (JSON-serialized) can be stored/retrieved by key, not just HostStatus. DataStore::get_value/set_value/set_value_if_absent provide the generic API; keyvalue rows are carried over when a temporary datastore is persisted, same as HostStatus.

persist()'s destination writes (HostStatus + every keyvalue row) are wrapped in a single transaction (write_persistent_data), rolling back on any failure -- a partial persist can no longer leave a datastore file that looks like a valid, fully-provisioned host on disk while reporting failure to its caller.

open() upgrades a pre-existing datastore that predates the keyvalue table, and rejects a truncated/schema-less file instead of silently treating it as a fresh, valid datastore.

No first-party consumer of the generic store in this PR: an earlier iteration added a database_id/datastore_id field as the first consumer, but the team decided it was effectively redundant with installation_id (added on top in #774) -- every real code path mints/reads both together -- so it was removed, leaving this PR as pure infrastructure. #774 is the first real consumer (installation_id/servicing_id).

Also includes two smaller, related fixes bundled into this PR:
* subsystems/management.rs: fix/unify the agent-config DatastorePath merge logic (a pre-existing bug in how a missing vs. present DatastorePath= line was handled).
* logging/tracestream.rs: let tests write the local metrics file to a temp path instead of the real host path, so datastore/tracestream tests can run as plain #[test]s instead of needing a VM.

Future application: trident-acl-agent uses a file (state.json) to track state, replace state.json with this datastore storage.

Related PRs in stack:
* [this PR] #773
* #774
* #778
* #779
* #781

### DIFF
--- a/crates/trident/src/datastore.rs
+++ b/crates/trident/src/datastore.rs
@@ -132,7 +132,7 @@ impl DataStore {
         db.execute(
             "CREATE TABLE IF NOT EXISTS hoststatus (
                 id INTEGER PRIMARY KEY AUTOINCREMENT,
-                timestamp DATETIME DEFALUT CURRENT_TIMESTAMP,
+                timestamp DATETIME DEFAULT CURRENT_TIMESTAMP,
                 contents TEXT NOT NULL
             )",
         )
@@ -185,11 +185,9 @@ impl DataStore {
             match query_statement.next() {
                 Ok(State::Done) => break,
                 Err(e) => {
-                    warn!(
-                        "Failed to get next keyvalue row while copying datastore: {:?}",
-                        e
-                    );
-                    break;
+                    return Err(e)
+                        .structured(ServicingError::from(DatastoreError::ReadDatastore))
+                        .message("Failed to get next keyvalue row while copying datastore");
                 }
                 Ok(State::Row) => {} // continue below
             }
--- a/crates/trident/src/datastore.rs
+++ b/crates/trident/src/datastore.rs
@@ -51,6 +51,11 @@ impl DataStore {
                 path: path.to_string_lossy().into(),
             },
         })?;
+        // Existing datastores may predate a table added in a later Trident
+        // version (e.g. `keyvalue`). Idempotently ensure the full schema is
+        // present so upgraded hosts don't fail with "no such table" the
+        // first time a new table is accessed.
+        Self::ensure_schema(&db)?;
         let host_status_yaml: Option<serde_yaml::Value> = db
             .prepare("SELECT contents FROM hoststatus ORDER BY id DESC LIMIT 1")
             .structured(ServicingError::Datastore {
@@ -129,6 +134,15 @@ impl DataStore {
 
         let db =
             sqlite::open(path).structured(ServicingError::from(DatastoreError::OpenDatastore))?;
+        Self::ensure_schema(&db)?;
+        Ok(db)
+    }
+
+    /// Idempotently create any tables that don't already exist. Safe to call
+    /// on both newly-created and pre-existing datastores, so that a
+    /// datastore created by an older Trident version picks up tables added
+    /// by a newer version the next time it is opened.
+    fn ensure_schema(db: &sqlite::Connection) -> Result<(), TridentError> {
         db.execute(
             "CREATE TABLE IF NOT EXISTS hoststatus (
                 id INTEGER PRIMARY KEY AUTOINCREMENT,
@@ -144,7 +158,7 @@ impl DataStore {
             )",
         )
         .structured(ServicingError::from(DatastoreError::InitializeDatastore))?;
-        Ok(db)
+        Ok(())
     }
 
     pub(crate) fn persist(&mut self, path: &Path) -> Result<(), TridentError> {
@@ -541,6 +555,36 @@ mod tests {
             datastore.get_value::<String>("some-key").unwrap(),
             Some("other-value".to_string())
         );
+
+        temp_dir.close().unwrap();
+    }
+
+    #[test]
+    /// Regression test: a datastore created by an older Trident version that
+    /// predates the `keyvalue` table (only `hoststatus` exists) must still
+    /// be usable after `open()` -- in particular, `correlation_id()` must
+    /// not fail with "no such table: keyvalue".
+    fn test_open_upgrades_pre_existing_datastore_schema() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let path = temp_dir.path().join("db.sqlite");
+
+        // Simulate a datastore created before the `keyvalue` table existed:
+        // create only the `hoststatus` table.
+        {
+            let db = sqlite::open(&path).unwrap();
+            db.execute(
+                "CREATE TABLE IF NOT EXISTS hoststatus (
+                    id INTEGER PRIMARY KEY AUTOINCREMENT,
+                    timestamp DATETIME DEFAULT CURRENT_TIMESTAMP,
+                    contents TEXT NOT NULL
+                )",
+            )
+            .unwrap();
+        }
+
+        let mut datastore = super::DataStore::open(&path).unwrap();
+        // Should not fail with "no such table: keyvalue".
+        datastore.correlation_id().unwrap();
 
         temp_dir.close().unwrap();
     }
--- a/crates/trident/src/datastore.rs
+++ b/crates/trident/src/datastore.rs
@@ -46,11 +46,15 @@ impl DataStore {
 
     pub(crate) fn open(path: &Path) -> Result<Self, TridentError> {
         debug!("Loading datastore from {}", path.display());
-        let db = sqlite::open(path).structured(ServicingError::Datastore {
+        let mut db = sqlite::open(path).structured(ServicingError::Datastore {
             inner: DatastoreError::LoadDatastore {
                 path: path.to_string_lossy().into(),
             },
         })?;
+        // Multiple connections to the same datastore file (e.g. concurrent
+        // daemon RPC handlers) can briefly contend for the write lock; wait
+        // for it rather than failing immediately with "database is locked".
+        Self::set_busy_timeout(&mut db)?;
         // Existing datastores may predate a table added in a later Trident
         // version (e.g. `keyvalue`). Idempotently ensure the full schema is
         // present so upgraded hosts don't fail with "no such table" the
@@ -132,10 +136,22 @@ impl DataStore {
             DatastoreError::CreateDatastoreDirectory,
         ))?;
 
-        let db =
+        let mut db =
             sqlite::open(path).structured(ServicingError::from(DatastoreError::OpenDatastore))?;
+        Self::set_busy_timeout(&mut db)?;
         Self::ensure_schema(&db)?;
         Ok(db)
+    }
+
+    /// Wait (rather than immediately failing with "database is locked") for
+    /// up to five seconds when another connection holds the write lock on
+    /// this datastore file. Needed because more than one connection to the
+    /// same datastore can legitimately exist at once (e.g. concurrent
+    /// daemon RPC handlers each calling `Trident::new`), and SQLite's
+    /// default busy timeout is zero.
+    fn set_busy_timeout(db: &mut sqlite::Connection) -> Result<(), TridentError> {
+        db.set_busy_timeout(5000)
+            .structured(ServicingError::from(DatastoreError::OpenDatastore))
     }
 
     /// Idempotently create any tables that don't already exist. Safe to call
@@ -376,26 +392,67 @@ impl DataStore {
     /// Values are serialized as JSON, so any type implementing
     /// `serde::Serialize`/`serde::de::DeserializeOwned` can be stored, not
     /// just `HostStatus`.
+    ///
+    /// `correlation_id` is currently the only first-party caller of the
+    /// generic key-value store, and it needs insert-if-absent semantics
+    /// (see `set_value_if_absent`) rather than an unconditional overwrite,
+    /// so this unconditional-overwrite variant is presently exercised only
+    /// by tests. It's kept as part of the generic key-value API (see
+    /// `get_value`) for future callers that do want overwrite semantics.
+    #[allow(dead_code)]
     pub(crate) fn set_value<T: Serialize>(&self, key: &str, value: &T) -> Result<(), TridentError> {
+        let contents = serde_json::to_string(value).structured(InternalError::SerializeValue {
+            key: key.to_string(),
+        })?;
+        self.write_key_value_row(
+            key,
+            &contents,
+            "INSERT INTO keyvalue (key, contents) VALUES (?, ?) \
+             ON CONFLICT(key) DO UPDATE SET contents = excluded.contents",
+        )
+    }
+
+    /// Like [`Self::set_value`], but only inserts a row if `key` does not
+    /// already have one; an existing row is left untouched. Used where two
+    /// datastore connections could race to perform "first access"
+    /// initialization of a key (see [`Self::correlation_id`]): whichever
+    /// connection's insert commits first wins, and the other's insert
+    /// becomes a no-op instead of overwriting the winner's value.
+    pub(crate) fn set_value_if_absent<T: Serialize>(
+        &self,
+        key: &str,
+        value: &T,
+    ) -> Result<(), TridentError> {
+        let contents = serde_json::to_string(value).structured(InternalError::SerializeValue {
+            key: key.to_string(),
+        })?;
+        self.write_key_value_row(
+            key,
+            &contents,
+            "INSERT INTO keyvalue (key, contents) VALUES (?, ?) \
+             ON CONFLICT(key) DO NOTHING",
+        )
+    }
+
+    /// Execute a parameterized `INSERT ... ON CONFLICT ...` against the
+    /// `keyvalue` table, binding `key` and `contents` as the two `?`
+    /// placeholders in `sql`.
+    fn write_key_value_row(
+        &self,
+        key: &str,
+        contents: &str,
+        sql: &str,
+    ) -> Result<(), TridentError> {
         let db = self
             .db
             .as_ref()
             .structured(ServicingError::from(DatastoreError::WriteToClosedDatastore))?;
 
-        let contents = serde_json::to_string(value).structured(InternalError::SerializeValue {
-            key: key.to_string(),
+        let mut statement = db.prepare(sql).structured(ServicingError::Datastore {
+            inner: DatastoreError::WriteKeyValue {
+                key: key.to_string(),
+            },
         })?;
-
-        let mut statement = db
-            .prepare(
-                "INSERT INTO keyvalue (key, contents) VALUES (?, ?) \
-                 ON CONFLICT(key) DO UPDATE SET contents = excluded.contents",
-            )
-            .structured(ServicingError::Datastore {
-                inner: DatastoreError::WriteKeyValue {
-                    key: key.to_string(),
-                },
-            })?;
         statement
             .bind((1, key))
             .structured(ServicingError::Datastore {
@@ -404,7 +461,7 @@ impl DataStore {
                 },
             })?;
         statement
-            .bind((2, &*contents))
+            .bind((2, contents))
             .structured(ServicingError::Datastore {
                 inner: DatastoreError::WriteKeyValue {
                     key: key.to_string(),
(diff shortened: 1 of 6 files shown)


crates/trident/src/datastore.rs
```diff
@@ -128,6 +137,13 @@ impl DataStore {
             )",
         )
         .structured(ServicingError::from(DatastoreError::InitializeDatastore))?;
+        db.execute(
+            "CREATE TABLE IF NOT EXISTS keyvalue (
+                key TEXT PRIMARY KEY,
+                contents TEXT NOT NULL
+            )",
```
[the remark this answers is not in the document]
bfjelds: Great catch — fixed the typo (DEFALUT -> DEFAULT) in the hoststatus schema so timestamp now auto-populates as intended. See 835e3dd6.

crates/trident/src/datastore.rs
```diff
@@ -139,13 +155,80 @@ impl DataStore {
...
+                Ok(State::Done) => break,
+                Err(e) => {
+                    warn!(
+                        "Failed to get next keyvalue row while copying datastore: {:?}",
+                        e
+                    );
+                    break;
+                }
+                Ok(State::Row) => {} // continue below
+            }
```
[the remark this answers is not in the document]
bfjelds: Great catch — copy_key_values now returns an error instead of warning and breaking out of the loop when a keyvalue row read fails, so persist() can no longer report success after a partial/failed copy. See 835e3dd6.

crates/trident/src/main.rs
```diff
@@ -168,6 +173,21 @@ fn run_trident(
                 let mut datastore = DataStore::open_or_create(agent_config.datastore_path())
                     .message("Failed to open datastore")?;
 
+                // Retrieve (or create, on first run) this host's unique
+                // correlation ID from the datastore, and attach it to all
+                // subsequent traces/metrics so they can be correlated with
+                // this specific installation.
+                match datastore.correlation_id() {
```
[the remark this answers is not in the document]
bfjelds: Great catch. Fixed by moving the correlation-ID retrieval into `Trident::new` itself, using the `datastore_path` it already receives, and setting it on the `TraceStream` before `trident_start` is emitted -- this covers every caller (the CLI path and each daemon gRPC service handler) uniformly, since they all supply `datastore_path`, without needing to duplicate the logic at each of the ~8 call sites. See f3eb5ab0.

One related question I'm leaving open rather than guessing at: `install()` (lib.rs) swaps in a fresh temporary datastore for multiboot installs before the new install is later persisted. That temporary datastore never inherits the outgoing datastore's correlation ID, so a multiboot reinstall on the same host currently gets a new correlation ID rather than carrying the old one forward. If that carryover is desired, it's a servicing-flow behavior change I'd rather scope separately -- happy to pick it up if you confirm the intended behavior.

crates/trident/src/datastore.rs
```diff
@@ -123,7 +132,14 @@ impl DataStore {
...
+                timestamp DATETIME DEFAULT CURRENT_TIMESTAMP,
+                contents TEXT NOT NULL
+            )",
+        )
+        .structured(ServicingError::from(DatastoreError::InitializeDatastore))?;
+        db.execute(
+            "CREATE TABLE IF NOT EXISTS keyvalue (
+                key TEXT PRIMARY KEY,
                 contents TEXT NOT NULL
             )",
```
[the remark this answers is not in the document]
bfjelds: Fixed. `open()` never ran the schema-creation statements that `make_datastore()` runs on first create, so an existing datastore that predates the `keyvalue` table would fail with "no such table: keyvalue" the first time `correlation_id()` (or any other keyvalue access) ran on it.

Extracted the `CREATE TABLE IF NOT EXISTS` statements into a shared idempotent `ensure_schema()` helper and call it from both `open()` and `make_datastore()`, so an upgraded datastore picks up any tables added by a newer Trident version the next time it's opened. Added a regression test (`test_open_upgrades_pre_existing_datastore_schema`) that creates a datastore with only the `hoststatus` table and verifies `open()` + `correlation_id()` succeed. See commit 17a2dd5b.

crates/trident/src/lib.rs
```diff
@@ -227,6 +227,23 @@ impl Trident {
             ));
         }
 
+        // Retrieve (or create, on first run) this host's unique
+        // correlation ID from the datastore actually used for servicing,
+        // and attach it to the shared TraceStream before any startup
+        // metrics are emitted, so every trace/metric -- including this
+        // very "trident_start" event -- carries it. This runs for every
```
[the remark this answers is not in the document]
bfjelds: Fixed. Moved the correlation ID retrieval/attach block earlier in `Trident::new`, before `hc.feature_tracing()` (and before any other tracing event), so the `host_config_feature_usage` metric now carries the correlation ID like every other trace/metric. See commit 17a2dd5b.

bfjelds:
Addressed two issues Copilot flagged in the review summary ("suppressed comments" section, not posted as separate inline threads):

1. **`datastore.rs` — concurrent first access to `correlation_id()` could return inconsistent IDs.** Fixed by replacing the read-then-write pattern with an atomic `INSERT ... ON CONFLICT DO NOTHING` (new `set_value_if_absent`) followed by a re-read, so racing connections converge on whichever UUID actually got persisted instead of silently overwriting each other. Also added a 5s SQLite busy timeout on connection open, since the atomic-insert path can hit `SQLITE_BUSY` under concurrent writers with the default 0ms timeout. Added a two-thread, barrier-synchronized regression test.

2. **`logging/tracestream.rs` — no regression test for correlation-ID enrichment.** Added a test asserting `set_correlation_id` is actually copied into `additional_fields` on emitted trace/metric entries, since the existing tests only asserted on `metric_name`/`value`.

Fixed in 6a70c74e.

crates/trident/src/datastore.rs
```diff
@@ -139,13 +185,78 @@ impl DataStore {
                 TridentVersion::SemVer(TRIDENT_SEMVER_VERSION.clone());
             Self::write_host_status(&persistent_db, self.host_status())?;
 
+            // Carry over any generic key-value entries (e.g. the correlation ID)
+            // recorded in the temporary datastore into the persistent one, so
+            // they survive the transition from temporary to persistent
+            // storage.
+            if let Some(temporary_db) = self.db.as_ref() {
+                Self::copy_key_values(temporary_db, &persistent_db)?;
+            }
```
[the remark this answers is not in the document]
bfjelds: Good catch. Fixed by fully reading and finalizing the source `SELECT` before issuing any writes to the destination, so the source's read lock is released before the destination write starts -- avoiding the same-file lock contention. Added a regression test that persists a temporary datastore to its own path. Fixed in 0e83655d.

crates/trident/src/logging/tracestream.rs
```diff
@@ -144,11 +177,16 @@ struct ExecutionTime(Instant);
 /// the tracing-subscriber crate to handle the events and send them to the
 /// server.
 impl TraceSender {
-    fn new(server: Arc<RwLock<Option<String>>>) -> Self {
+    fn new(
+        server: Arc<RwLock<Option<String>>>,
+        correlation_id: Arc<RwLock<Option<String>>>,
+        metrics_file_path: &str,
```
bfjelds, line 183: should this be a path obj

crates/trident/src/datastore.rs
```diff
@@ -37,11 +46,20 @@ impl DataStore {
...
         })?;
+        // Multiple connections to the same datastore file (e.g. concurrent
+        // daemon RPC handlers) can briefly contend for the write lock; wait
+        // for it rather than failing immediately with "database is locked".
+        Self::set_busy_timeout(&mut db)?;
+        // Existing datastores may predate a table added in a later Trident
+        // version (e.g. `keyvalue`). Idempotently ensure the full schema is
+        // present so upgraded hosts don't fail with "no such table" the
+        // first time a new table is accessed.
+        Self::ensure_schema(&db)?;
```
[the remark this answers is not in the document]
bfjelds: Great catch, thanks! Fixed in 11809790: `open()` now checks that the `hoststatus` table actually exists (via a new `table_exists()` helper) before treating the file as a valid existing datastore, instead of unconditionally running `CREATE TABLE IF NOT EXISTS hoststatus` first. A truncated/zero-byte or schema-less file is now rejected instead of silently opened as an empty temporary datastore. Added a regression test covering the truncated-file case.

crates/trident/src/main.rs
```diff
@@ -165,6 +165,10 @@ fn run_trident(
                 )
                 .message("Failed to initialize Trident")?;
 
+                // `Trident::new` has already retrieved (or created) this
+                // host's persisted database ID and attached it to the
+                // shared TraceStream, so every trace/metric emitted from
+                // here on -- including "trident_start" -- carries it.
                 let mut datastore = DataStore::open_or_create(agent_config.datastore_path())
```
frhuelsz, line 168: my take is to move this up before `Trident::new()` and init/obtain the install ID at that point, and setting it up against the tracer. @fintelia thoughts?

fintelia: I think the reason that we have things in this order is because opening the datastore might generate trace events

frhuelsz: hmm there is already a `if let Ok() = DataStore::open(datastore_path)` in `lib.rs:150` To extract a single thing. Perhaps we could replicate the pattern to try to extract the datastore UUID eagerly. If missing we create a new one, and make sure it gets passed onto the newly created datastore. We also need to mind the upgrade from ephemeral to persistent datastore.

bfjelds: i think moving it up before Trident::new means that there are error cases where we'd end up creating a datastore that we previously wouldn't have.

frhuelsz: my thoughts: Trident::new should also return the database object it opened/created inside, and avoid re-opening. The caller can just drop it when not needed, but i presume almost all callers of trident::new immediately open the datastore after this.

bfjelds: looking more into this, removing datastore_id allows us to remove the Trident::new() datastore creation path.

crates/trident/src/datastore.rs
```diff
@@ -139,13 +212,97 @@ impl DataStore {
                 TridentVersion::SemVer(TRIDENT_SEMVER_VERSION.clone());
             Self::write_host_status(&persistent_db, self.host_status())?;
 
+            // Carry over any generic key-value entries (e.g. an installation
+            // ID) recorded in the temporary datastore into the persistent
+            // one, so they survive the transition from temporary to
+            // persistent storage.
+            if let Some(temporary_db) = self.db.as_ref() {
+                Self::copy_key_values(temporary_db, &persistent_db)?;
```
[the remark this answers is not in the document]
bfjelds: Great catch. Fixed by wrapping the HostStatus write and every key-value insert in a single destination transaction (BEGIN/COMMIT, with ROLLBACK on any failure), so a failure partway through -- e.g. a key-value insert error after the HostStatus row already autocommitted -- no longer leaves a half-written file that reopens as a seemingly-valid, fully-provisioned datastore. Split copy_key_values() into read_key_values()/write_key_values() to make this possible. Added test_persist_rolls_back_on_partial_failure, which forces a write failure mid-transaction (via PRAGMA query_only) and asserts the destination file has no committed hoststatus row afterward. See 66a0dd9e.

crates/trident/src/datastore.rs
```diff
@@ -137,7 +210,48 @@ impl DataStore {
...
+            // false -- autocommitted on its own, so a later `open()` would
+            // treat this half-written file as a normal, fully-provisioned,
+            // non-temporary datastore even though this call reports
+            // failure to its caller and no key-value data (or only a
+            // partial prefix of it) actually made it across.
+            persistent_db
+                .execute("BEGIN")
+                .structured(ServicingError::from(DatastoreError::WriteToDatastore))?;
+            let result = Self::write_host_status(&persistent_db, self.host_status())
+                .and_then(|()| Self::write_key_values(&persistent_db, &key_values));
```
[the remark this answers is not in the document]
bfjelds: Great catch. Extracted the BEGIN/write_host_status/write_key_values/COMMIT-or-ROLLBACK block into its own write_persistent_data() function that persist() calls, and rewrote the test to call write_persistent_data() directly instead of re-implementing its transaction control.

Also caught a second, subtler gap in the process: my first attempt forced the failure via PRAGMA query_only, which blocks every write on the connection -- including the first one (write_host_status) -- so it could never actually distinguish an already-succeeded write being rolled back from that write never having succeeded in the first place. Switched to a trigger that fails only one specific key-value insert (RAISE(ABORT)), so write_host_status genuinely succeeds before write_key_values fails. Verified locally by temporarily stripping the transaction wrapping and confirming the test now fails as expected. See 9080d8b4.

crates/trident_api/src/error.rs
```diff
@@ -124,6 +124,12 @@ pub enum InternalError {
     #[error("Failed to serialize Host Status")]
     SerializeHostStatus,
 
+    #[error("Failed to serialize value for datastore key '{key}'")]
+    SerializeValue { key: String },
```
fintelia, line 128: This field name should mention datastore somewhere. `InternalError::SerializeValue` is a little too broad sounding giving the error string.

fintelia: Or maybe this just needs to be moved to the `DatastoreError` enum?

fintelia (dismissed):

crates/trident/src/datastore.rs
```diff
@@ -137,7 +210,20 @@ impl DataStore {
...
+            // out of the temporary datastore before starting any
+            // destination writes below: this both avoids a same-file
+            // read/write lock deadlock when `path` is the datastore's own
+            // current path (see `write_key_values`'s doc comment) and lets
+            // the HostStatus write and every key-value insert run inside a
+            // single destination transaction.
+            let key_values = match self.db.as_ref() {
+                Some(temporary_db) => Self::read_key_values(temporary_db)?,
+                None => Vec::new(),
+            };
```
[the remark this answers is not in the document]
bfjelds: Great catch — fixed in afa56ec9: `persist()` now returns `DatastoreError::WriteToClosedDatastore` instead of silently treating a closed temporary datastore as having zero key-value rows, and a regression test (`test_persist_on_closed_datastore_fails`) now covers this.

crates/trident/src/subsystems/management.rs
```diff
@@ -92,43 +92,67 @@ fn configure_agent_config(
...
         }
 
-        let datastore_configuration = format!("DatastorePath={}", datastore_path.display());
-        fs::write(agent_config_path, datastore_configuration).structured(
+        let mut updated_contents = contents;
+        if !updated_contents.is_empty() && !updated_contents.ends_with('\n') {
+            updated_contents.push('\n');
+        }
+        updated_contents.push_str(&format!("DatastorePath={}\n", datastore_path.display()));
+        fs::write(agent_config_path, updated_contents).structured(
```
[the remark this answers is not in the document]
bfjelds: Great catch, and you're right that this reintroduces the non-atomic write risk — that's intentional. This line previously used `atomic_write_file` (see the prior commit), but that was reverted deliberately after a separate review flagged that atomic (rename-based) replacement drops the destination file's SELinux context, since the SELinux subsystem only relabels on clean-install/AB-update, not other servicing types or stream-image (see #discussion_r4064492867).

We evaluated both risks together:
- This write only ever fires once per filesystem instance: the first time a non-default `DatastorePath=` needs merging into the agent config. Every subsequent `configure_agent_config` call for that filesystem sees the line already present and takes the read-only comparison branch — no further writes happen.
- That first write overwhelmingly coincides with clean-install or AB-update (the two servicing types that provision a fresh filesystem), which is exactly when the SELinux subsystem does relabel — so atomic writes and label loss weren't even actually safe for the common case in isolation; the atomic fix would've been fixing a rare corruption window while opening a comparably-scoped labeling window with no corresponding SELinux fix.
- Given both are narrow, low-frequency risks tied to the same one-time write, we chose to keep the code path simple (`fs::write`) and accept the original corruption risk rather than layer on `atomic_write_file` + a manual SELinux-context restore/`restorecon` step to close the gap it reopens.

No further code change planned here — happy to revisit if the frequency assumptions above turn out to be wrong (e.g. if there are servicing paths that re-trigger this write outside clean-install/AB-update more often than expected).

crates/trident/src/subsystems/management.rs
```diff
@@ -92,43 +92,67 @@ fn configure_agent_config(
...
-                return Err(TridentError::new(
-                    InvalidInputError::ImageBadAgentConfiguration,
-                ))
-                .message(format!(
-                    "Datastore path in agent config ({}) does not match expected path ({})",
-                    datastore_path_configured,
-                    datastore_path.display()
-                ));
-            }
+    // Ensure that Trident agent config exists with correct datastore path.
```
frhuelsz: I'm not entirely sure I understand why this function was changed or how this is related.

crates/trident/src/datastore.rs
```diff
@@ -118,26 +146,90 @@ impl DataStore {
...
             sqlite::open(path).structured(ServicingError::from(DatastoreError::OpenDatastore))?;
+        Self::set_busy_timeout(&mut db)?;
+        Self::ensure_schema(&db)?;
+        Ok(db)
+    }
+
+    /// Wait (rather than immediately failing with "database is locked") for
+    /// up to five seconds when another connection holds the write lock on
+    /// this datastore file. Needed because more than one connection to the
+    /// same datastore can legitimately exist at once (e.g. concurrent
```
frhuelsz, line 159: hmm this is mostly false, grpc imposes 2 levels of locks to avoid concurrency, the only possible way this happens is if you do grpc and old-cli at the same time, which is bad and we probably want to fail on anyways.

crates/trident/src/datastore.rs
```diff
@@ -118,26 +146,90 @@ impl DataStore {
             DatastoreError::CreateDatastoreDirectory,
         ))?;
 
-        let db =
+        let mut db =
             sqlite::open(path).structured(ServicingError::from(DatastoreError::OpenDatastore))?;
+        Self::set_busy_timeout(&mut db)?;
+        Self::ensure_schema(&db)?;
```
frhuelsz, line 152: curious on why ensure_schema is run twice?

crates/trident/src/subsystems/management.rs
bfjelds: the change belongs in the next PR ... it is part of handling an AgentConfig that set Telemetry=OptIn only, which the previous behavior failed.

thanks for catching.  i will move to pr 774.

crates/trident/src/datastore.rs
bfjelds: sorry, i'm missing what you're seeing.  i see it called once during make_database and once during open.  can you clarify?

crates/trident/src/datastore.rs
bfjelds: lol ... good to konw.  this came from github's copilot worrying about concurrent requests causing simultaneous installation id creation requests.

crates/trident/src/subsystems/management.rs
bfjelds: done

crates/trident/src/datastore.rs
bfjelds: is there harm in the extra protection here?  happy to remove given the protections elsewhere.

crates/trident/src/logging/tracestream.rs
```diff
@@ -127,7 +127,20 @@ impl TraceStream {
...
+        self.make_trace_sender_with_metrics_path(TRIDENT_METRICS_FILE_PATH)
+    }
+
+    /// Like `make_trace_sender`, but writes the local metrics file to
+    /// `metrics_file_path` instead of the real host path
+    /// (`TRIDENT_METRICS_FILE_PATH`). This lets tests exercise the full
+    /// metrics-writing pipeline against a throwaway temp file instead of a
+    /// real, shared host path, so they can be plain `#[test]`s instead of
+    /// needing a VM.
+    pub(crate) fn make_trace_sender_with_metrics_path(
```
frhuelsz: nit doesn't need to be pub at all i think?

crates/trident/src/logging/tracestream.rs
```diff
@@ -164,6 +177,12 @@ impl TraceSender {
         self.server.read().map(|s| s.clone()).unwrap_or_default()
     }
 
+    /// Build the `additional_fields` map for a trace entry from the static
+    /// `ADDITIONAL_FIELDS`.
+    fn additional_fields(&self) -> BTreeMap<String, Value> {
```
frhuelsz: is this fn needed?

crates/trident/src/logging/tracestream.rs
bfjelds: Good catch \xe2\x80\x94 fixed in 5885e77e, dropped the `pub(crate)` since this fn is only used within this file (impl block + the `#[cfg(test)] mod tests` submodule).

crates/trident/src/logging/tracestream.rs
bfjelds: Good catch \xe2\x80\x94 fixed in 5885e77e, removed the wrapper and inlined `ADDITIONAL_FIELDS.clone()` directly at both call sites.

crates/trident/src/main.rs
bfjelds: Following up on where this landed. Rather than reordering `Trident::new()` relative to datastore open, the underlying concern (early trace events emitted without the persisted install ID) was resolved via a pre-warm step added in #774: `main.rs` now attaches `installation_id`/`database_id` to the shared `TraceStream` before entering `run_with_operation`, mirroring the existing daemon-startup pre-warm, so `command_start`/`trident_start` carry the ID for both the CLI and daemon paths uniformly.

`Trident::new()` itself still opens/creates the datastore internally as before -- we did not pursue having it return the opened datastore object to the caller or removing its datastore-creation path, so that specific architectural change discussed here was not implemented. Resolving since the practical telemetry-attribution gap is closed; happy to reopen if the caller-owns-the-datastore refactor is still wanted.

crates/trident_api/src/error.rs
bfjelds: Great catch — this is already done. `SerializeValue` (along with `DeserializeValue`, `ReadKeyValue`, etc.) now lives in the `DatastoreError` enum rather than `InternalError`, so the variant name no longer needs to mention "datastore" itself since it's scoped there. Resolving.

PR #789, opened 2026-09-09
[RFC] feature: Carry sysext and confext images in COSI files

frhuelsz:
## What

Adds an RFC proposing that COSI files carry sysext and confext images.

Extension images become ZSTD-compressed members of the COSI tar under
`images/extensions/`, alongside partition images, and a new optional
`extensions` section in the COSI metadata describes them. Each entry reuses the
existing `ImageFile` object and carries an optional destination path, mirroring
the `Extension` object in the Host Configuration.

## Why

An extension is currently a second artefact with a separate lifecycle: hosted
separately, fetched from a second endpoint at deploy time, with integrity
anchored outside `os.image.sha384`, and updated through a Runtime Update rather
than the A/B update that ships the OS.

Bundling delivers and activates operator content in the same A/B update and the
same reboot as the OS, reusing the existing rollout, health-gating and rollback
machinery. Rollback needs no additional work, since extension images are files
in the target slot's filesystem and the previous slot retains the previous set.

This adds an option rather than replacing one. Extensions deliberately versioned
independently of the OS should continue to use `os.sysexts` and `os.confexts`.

## Notable decisions

- **Two arrays (`sysexts`, `confexts`) rather than one with a `kind` field.**
  The kinds have disjoint permitted directories, defaults, `extension-release`
  locations, identity fields and activation units, and this is the shape the
  Host Configuration already uses.
- **Payloads stay under `images/`.** The specification already permits
  subdirectories, so `ImageFile.path`'s `^images/.+` pattern does not need
  relaxing. Relaxing it would produce a 1.3 `ImageFile` that fails the 1.0-1.2
  schemas.
- **Entries carry only `image` and an optional `path`.** Name, kind, extension
  ID and compatibility metadata are all derivable from the payload.
- **Bundled and Host Configuration extensions are merged; a collision is an
  error.** Both lists are legitimate at once, and silent precedence would run
  software the image author did not validate, with no visible symptom.
- **`ID=_any` is not warned on or refused.** It describes compatibility, not
  delivery. The useful check is the inverse: a bundled `ID=<distro>` extension
  that does not match the COSI's own `osRelease`.
- **SELinux support is unchanged**, but the enforcing-mode check must read the
  merged set rather than the Host Configuration alone.

Two integration points are easy to miss and are called out in the RFC:
`osconfig` enables `systemd-sysext.service` and `systemd-confext.service` based
on the Host Configuration lists being non-empty, and the SELinux dynamic
validation reads the same lists. Both must read the merged set, or a COSI-only
extension configuration is silently unmerged or silently mislabelled.

## Open questions

1. Should an override escape hatch exist? The RFC makes a Host Configuration
   versus COSI collision an error. An explicit opt-in on the Host Configuration
   side would permit pinning a hotfixed extension, and equally permit running a
   combination the image author did not validate.
2. Should a bundled extension be suppressible? A deny list has no analogue in
   the current API and no concrete requester.
3. Should `path` be required? It is optional here for symmetry with
   `Extension.path`.
4. `extensions` containing `sysexts` and `confexts`, or two top-level arrays?
5. Should Host Configuration extension ID uniqueness be enforced independently
   of this work? It is documented but unenforced today.
6. Should writers be required to place extension payloads after all region
   images, or merely encouraged? The RFC says SHOULD.

## Known limitation

An existing Trident accepts any `1.x` COSI and ignores unknown fields, so it
will deploy a 1.3 COSI correctly and omit the extensions without reporting
anything. This cannot be corrected in-band for readers that already exist. The
RFC proposes documenting it, warning on unknown minor versions from now on, and
health-check gating where the target's Trident version cannot be controlled.

### DIFF
--- a/docs/Development/RFCs/0000-cosi-extension-images.md
+++ b/docs/Development/RFCs/0000-cosi-extension-images.md
@@ -1,0 +1,774 @@
+# 0000 COSI Extension Images
+
+- Date: 2026-08-28
+- RFC PR: [microsoft/trident#0000](https://github.com/microsoft/trident/pull/0000)
+- Issue: [microsoft/trident#0000](https://github.com/microsoft/trident/issues/0000)
+
+## Summary
+
+This RFC proposes carrying systemd system extension (sysext) and configuration
+extension (confext) images inside a COSI file. Extension images become ZSTD
+compressed members of the COSI tar, exactly like partition images, and a new
+optional `extensions` section in `metadata.json` describes them. Each entry
+reuses the existing [`ImageFile`](../../Reference/Composable-OS-Image.md#imagefile-object)
+object to point at the payload in the tar, and carries the destination path on
+the target OS, mirroring the existing
+[`Extension`](../../Reference/Host-Configuration/API-Reference/Extension.md)
+object in the Host Configuration. Trident then deploys those extensions as part
+of the image, in the same operation and the same reboot as the rest of the OS,
+instead of fetching them from a separate location at deploy time.
+
+## Motivation and Goals
+
+Today the only way to get an extension onto a Trident-managed host is
+`os.sysexts` / `os.confexts` in the Host Configuration, where each entry is a
+URL plus a SHA-384. That works, but it makes the extension a second artefact
+with a second lifecycle:
+
+- **A second thing to host.** The operator must publish the `.raw` DDI
+  somewhere Trident can reach (`http://`, `https://`, `file://` or `oci://`)
+  and keep it there for as long as any host might re-run the flow.
+- **A second download at deploy time.** The COSI is streamed sparsely from one
+  endpoint; each extension is fetched in full from another. Two endpoints means
+  two sets of network, proxy, TLS and authentication failure modes on the
+  provisioning path.
+- **A second integrity story.** The COSI's integrity is anchored by the
+  `metadata.json` hash in `os.image.sha384`, which transitively covers every
+  partition image. An extension's hash is anchored by whoever wrote the Host
+  Configuration. Bundling puts the extension under the same root of trust as
+  the rest of the OS.
+- **A second update.** This is the important one. Extension changes are
+  currently a Runtime Update, distinct from the A/B update that ships the OS.
+  Delivering operator content that must land alongside a new OS therefore means
+  two operations, two rollout gates and potentially two disruptions.
+
+### A/B Updates Are the Strategic Point
+
+If extensions ride the COSI, then operator content is delivered and activated
+in the *same* A/B update and the *same* reboot as the OS. It reuses the
+rollout, [health-gating](../../Reference/Host-Configuration/API-Reference/Health.md)
+and rollback machinery that already exists for images, rather than needing a
+parallel channel with its own staging, verification and failure semantics.
+
+Rollback becomes trivially correct: extension images are files inside the
+target slot's own filesystem, so the previous slot still holds the previous
+extension set, untouched. Rolling back to the previous root volume rolls back
+the extensions with it. See [Rollback](#rollback).
+
+### Why Not Just Bake the Files Into the Root Filesystem Image?
+
+An obvious objection: the image builder controls the root filesystem image, so
+it could simply place `my-tool.raw` in `/var/lib/extensions/` before the COSI
+is built, with no spec change at all. That is a genuine alternative and it
+works for some configurations. It loses on four counts:
+
+1. **Verity.** With root or usr verity, any change to the protected filesystem
+   changes the root hash and requires re-signing. Extensions live outside the
+   protected tree precisely so they can change independently; the mechanism
+   that makes them attractive is defeated if adding one means rebuilding and
+   re-signing the base image.
+2. **Destination volumes that are not image-provisioned.** `/var/lib/extensions/`
+   is commonly on a separate volume, which may be newly created rather than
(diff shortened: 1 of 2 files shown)


frhuelsz:
Added a section on extension-only updates, and revised the scope accordingly.

**The gap was real.** `select_servicing_type` short-circuits: `ab_update_required()` returns true whenever `os.image.sha384` differs, before any subsystem is consulted. The metadata hash covers the `extensions` section, so with the RFC as originally written, changing only a bundled extension would have forced a full A/B update and a reboot. That made bundling strictly more disruptive than the current Host Configuration route.

**An extension-only COSI is the wrong shape.** The metadata root requires `images`, `disk` with at least one `gptRegions` entry, `bootloader` and `osPackages`, so a COSI without partition images is not representable, and making it so would require those fields to become optional. Every consumer of `os.image` — filesystem source resolution, `derive_host_configuration`, ESP detection, verity setup — assumes the image describes a complete OS. The distinction is drawn on content equality instead.

**Mechanism.** An OS content digest over the parts of the metadata that determine the deployed OS (`images[]` and `disk.gptRegions[]` hashes, `disk` geometry, `bootloader`, `osArch`, `osRelease`), excluding `extensions`, recorded in the Host Status. Two COSIs with equal digests deploy an identical OS and differ only in their extensions. Trident cannot determine this otherwise: it retains the URL and metadata hash of the applied image, not the metadata, and re-fetching the previous COSI to diff it is not dependable.

The path is requested and then verified, never inferred. It proceeds only when a Runtime Update is explicitly requested, the digests match, and the recorded image is the deployed one. Otherwise Trident refuses with a structured error naming what differs. It must not promote the operation to an A/B update, and must not apply the extension change while leaving other changes unapplied. Refusal is the only acceptable failure, which is what keeps a partially-applied state unreachable.

**Rollback is weaker on this path**, and that is why it is not automatic. `set_up_extensions` skips removal of the superseded image only on Clean Install and A/B Update, so a Runtime Update replaces files in the running slot. Rollback re-runs the subsystem with the specs reversed and re-acquires the previous extension image from its source, which for a bundled extension is the previous COSI. A runtime rollback therefore requires the previous COSI to remain reachable; an A/B rollback requires nothing. That trade should be made by the operator, not inferred from a property of two images.

Three questions added to the open list:

- What surface requests an extension-only Runtime Update? There is none today; `forceAbUpdate` is the nearest precedent.
- What exactly does the OS content digest cover? Excluding too much permits a Runtime Update between images differing in ways Trident does not model; including too much rejects legitimate cases.
- Should the digest be published in the COSI metadata rather than computed? Computing it needs no writer cooperation and cannot be misdeclared, but publishing would let a writer state intent directly.

frhuelsz:
Revised, on both points.

**Storing the metadata.** Replaced the single digest with a structured summary of the deployed image's metadata, recorded in the Host Status: for each entry in `images[]` and `disk.gptRegions[]`, the `image.sha384`, `uncompressedSize` and the entry's identity (partition number, mount point, `fsUuid`, `partType`, verity root hash); plus `osArch`, `osRelease`, `disk` geometry and `bootloader`.

Excluded, with reasons: `extensions` (the subject of the comparison), `osPackages` (validated but never acted on), `compression.maxWindowLog` (governs decompression, and a new extension may legitimately raise it), and `id` (identifies the file, not its content).

A record rather than a digest, since the digest is derivable from it and the record additionally lets Trident report *which* partition differs. Not the full metadata, since `osPackages` is the bulk of it and has no bearing on the decision.

This gives the three outcomes, including the no-op case, which the previous revision missed:

| Summary | Extension set | Result |
| --- | --- | --- |
| Equal | Equal | No-op. |
| Equal | Differs | Extension-only change; a Runtime Update suffices. |
| Differs | Any | A/B update. |

`os.image.sha384` changes whenever any part of the metadata changes, including `id` and the ordering of `osPackages`, so a materially identical image currently forces an A/B update.

The summary carries its own schema version. Where it is absent or of an unrecognised version, Trident falls back to comparing `os.image.sha384`. That confines a change to `ab_update_required()` — a core decision — to hosts that have a recorded summary, and gives existing hosts an unchanged upgrade path.

**Construction, and the reproducibility problem.** This is the more important of the two, and it invalidates the obvious approach. Filesystem UUIDs, inode timestamps, superblock creation and mount times and allocation order all vary between builds, and ZSTD output varies with compression level and library version. Rebuilding from the same source will not generally produce the same images.

So requiring reproducible rebuilds is the wrong approach. The build operation is not "rebuild the image identically with a different extension set", it is **"copy an existing COSI, replacing its extension set"**. Region images and their metadata entries are copied verbatim from the input COSI; no filesystem is rebuilt and no region image is recompressed. Only the extension tar members, the `extensions` section, `compression.maxWindowLog` where the new extensions need a larger window, and `id` change.

Equality is then guaranteed by construction rather than hoped for through reproducibility, and the operation is a tar rewrite rather than an image build. This reframes the writer requirement from a reproducible-builds effort into a bounded COSI-to-COSI rewrite, and it is now listed under Dependencies.

Where a COSI is rebuilt from source instead, its region images differ, the summary differs, and Trident selects an A/B update. That is the safe outcome: the mechanism degrades to current behaviour rather than misclassifying a changed OS as unchanged. **The failure mode of an unreproducible build is a redundant A/B update, never a skipped one.**

Open questions updated accordingly: the exact coverage of the summary (`version` is currently excluded), record against digest, and whether the no-op outcome belongs in this RFC at all, given it is a behavioural change independent of extensions.

PR #791, opened 2026-09-11
ci: enable Copilot cloud agent on 1ES

Britel:
## Overview

This gives GitHub Copilot coding agents a reproducible Trident environment where they can make changes, build the project, and run unit tests before opening a PR.

## Changes

- Run Copilot on Trident's ephemeral 1ES runner scale set.
- Install the documented native, Go, Rust, and protobuf dependencies.
- Authenticate to Azure DevOps with the runner's managed identity and use Trident's internal Cargo feed.
- Keep Cargo output on `/mnt/storage` while preserving the repository's expected `target/` paths.
- Restrict privileged setup validation to trusted pushes to `main`.

## Validation

- Build and unit tests: https://github.com/microsoft/trident/actions/runs/34640841874
  - `make build` succeeded
  - `make test` passed 1,057 tests with 0 failures and 3 ignored
- Managed identity: https://github.com/microsoft/trident/actions/runs/35038952945
  - fetched Cargo dependencies through `BMP_PublicPackages`
  - downloaded `go-tools` from ECF pipeline 5067

Functional and E2E test execution remain follow-up work.


scripts/cloud-agent/setup.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+readonly RUST_TOOLCHAIN="1.93.0"
```
bfjelds: is this something we have to maintain?  if so, can it read the rust version from one of the Cargo files?

scripts/cloud-agent/setup.sh
```diff
@@ -0,0 +1,80 @@
...
+readonly PROTOC_VERSION="33.2"
+readonly PROTOC_SHA256="b24b53f87c151bfd48b112fe4c3a6e6574e5198874f38036aff41df3456b8caf"
+readonly PROTOC_GEN_GO_VERSION="v1.36.11"
```
bfjelds, line 6: similar question here ... maybe we can use the Makefile target here to install protoc?

scripts/cloud-agent/setup.sh
fintelia: Is there any problem with using 1.98.1 here? We're sometimes slow to bump versions, but if we're to specify a specific version we might as well not start with something that's 8 months out of date :smile:

fintelia (dismissed):

scripts/cloud-agent/setup.sh
```diff
@@ -0,0 +1,80 @@
...
+    if [[ ! -d target || -n "$(find target -mindepth 1 -maxdepth 1 -print -quit)" ]]; then
+        echo "Refusing to replace existing non-empty target path" >&2
+        exit 1
+    fi
+    rmdir target
+fi
+ln -sfn "$CARGO_TARGET_PATH" target
+test "$(readlink -f target)" = "$CARGO_TARGET_PATH"
+
+cargo fetch --locked
```
frhuelsz, line 69: heads up: this `cargo fetch` needs azure devops auth to work at all. our `.cargo/config.toml` replaces crates.io with the private ADO feeds (BMP_PublicPackages, etc), and `make .cargo/config` here runs without `OVERRIDE_RUST_FEED=true`, so `replace-with = "BMP_PublicPackages"` stays on and *every* crate pull goes through them, not just internal ones.

but this workflow injects no secrets and only has `contents: read`, so nothing here provides that auth — it'd only work if the 1ES image bakes in a managed identity / cred provider. the validation that showed build+test passing used the smoke workflow that's since been removed, so it doesn't prove the copilot agent's own job context gets the same creds. can you confirm the agent image actually authenticates to these feeds? if not, `cargo fetch` fails during setup, copilot starts with a broken env, and every build in the session fails — silent and confusing.

scripts/cloud-agent/setup.sh
Britel: The Cargo `rust-version` is Trident's MSRV (currently 1.85.0), not the development toolchain version. The setup now pins Rust 1.98.1 explicitly, matching the latest stable version exercised by the setup validation run.

Britel: No blocker. The setup now pins Rust 1.98.1 explicitly.

scripts/cloud-agent/setup.sh
Britel: There was not an existing Make target, so I added `make install-protoc` and changed the setup script to use it. The target installs protoc 33.2 with verified checksums for both x86_64 and aarch_64.

scripts/cloud-agent/setup.sh
Britel: Good catch on making this explicit. The Makefile already defaulted `OVERRIDE_RUST_FEED=true`, so the successful setup run used public crates.io rather than the ADO replacement feed. Commit 8781c1d now forces regeneration with `make -B OVERRIDE_RUST_FEED=true .cargo/config` and fails setup if an active `replace-with` remains, so the cloud agent does not depend on ADO feed credentials.

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,35 @@
...
+permissions: {}
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - Makefile
+      - .github/workflows/copilot-setup-steps.yml
+      - scripts/cloud-agent/setup.sh
```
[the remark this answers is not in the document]
Britel: `copilot-setup-steps.yml` is special configuration consumed by the Copilot coding agent; agent sessions do not start it through `workflow_dispatch`. GitHub runs the single `copilot-setup-steps` job before the agent begins, provided the file exists on the default branch. The `on` block only controls ordinary Actions validation of changes to this file. Manual dispatch was intentionally removed because it allowed a selected feature-branch ref to run privileged setup code on the self-hosted runner.

GitHub documents the special invocation here: https://docs.github.com/en/copilot/customizing-copilot/customizing-the-development-environment-for-copilot-coding-agent

scripts/cloud-agent/setup.sh
```diff
@@ -0,0 +1,88 @@
...
+    >/dev/null
+
+install -D -m 0755 \
+    scripts/cloud-agent/ado-cargo-token.sh \
+    "$CARGO_CREDENTIAL_PROVIDER_PATH"
+
+make -B .cargo/config
+printf '\n[registry]\nglobal-credential-providers = ["cargo:token-from-stdout %s"]\n' \
+    "$CARGO_CREDENTIAL_PROVIDER_PATH" \
+    >> .cargo/config
```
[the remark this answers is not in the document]
Britel: Good catch that the production setup was not actually exercising the private feed. The cause was the Makefile's `OVERRIDE_RUST_FEED ?= true` default, not the credential-provider syntax. Commit `b3045f3` now calls `make -B OVERRIDE_RUST_FEED=false .cargo/config` and asserts that `replace-with = "BMP_PublicPackages"` remains active.

`cargo:token-from-stdout` accepts the command directly without a `--` separator, and the same provider string fetched through `BMP_PublicPackages` successfully in run https://github.com/microsoft/trident/actions/runs/35038952945.

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - Makefile
+      - .github/workflows/copilot-setup-steps.yml
+      - scripts/cloud-agent/ado-cargo-token.sh
+      - scripts/cloud-agent/setup.sh
```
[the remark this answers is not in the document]
Britel: Addressed in `d7aee34`. The trusted setup validation now also runs when `.cargo/config.toml`, `Cargo.lock`, or any workspace `Cargo.toml` changes, covering the direct inputs to the internal-feed `cargo fetch --locked` validation.

Britel:
@frhuelsz Added a UMI so the job can reach BMP_PublicPackages. All concerns should be addressed.

Perhaps as a follow-up we could consolidate some of these pined versions across the infra.

frhuelsz (approved):

PR #794, opened 2026-09-14
feature: to enable ACL + IPE, copy .cred files from UKI addon dirs to ESP

bfjelds:
systemd-stub consumes both `*.addon.efi` and `*.cred` files from a UKI's `.extra.d/` directory at boot. `stage_uki_on_esp` only copied `*.addon.efi` files, silently dropping any `*.cred` files, so encrypted credentials placed in a COSI's `extra.d` were never installed or updated onto the ESP.

## Change
Extend the addon-directory copy filter in `stage_uki_on_esp` to also accept the `.cred` suffix (new `UKI_CRED_FILE_SUFFIX` constant in `osutils::uki`), so credential files are copied to the ESP alongside addon files on both install and update.

## Testing
- `cargo test -p trident --lib engine::boot::uki::` - 36 passed
- `cargo test -p osutils --lib uki::` - 8 passed
- `cargo fmt --check` - clean
- `make target/release/trident` - builds successfully


crates/trident/src/engine/boot/uki.rs
```diff
@@ -155,13 +157,12 @@ pub fn stage_uki_on_esp(
             continue;
         }
 
-        if !path
-            .as_os_str()
-            .as_encoded_bytes()
-            .ends_with(UKI_ADDON_FILE_SUFFIX.as_bytes())
+        let file_name_bytes = path.as_os_str().as_encoded_bytes();
+        if !file_name_bytes.ends_with(UKI_ADDON_FILE_SUFFIX.as_bytes())
+            && !file_name_bytes.ends_with(UKI_CRED_FILE_SUFFIX.as_bytes())
```
bfjelds, line 162: should this be ACL only?

frhuelsz, line 162: While we're in here, it might be worth extending the filter to cover the *full* set of files systemd-stub reads from a UKI's `.extra.d/`, rather than adding them one feature at a time. Per [systemd-stub(7)](https://www.freedesktop.org/software/systemd/man/latest/systemd-stub.html), that directory can hold four kinds of drop-ins: `*.addon.efi` (PE addons) and `*.cred` (credentials) — both handled after this change — plus `*.raw`/`*.sysext.raw` (system-extension images, unpacked to `/.extra/sysext/`) and `*.confext.raw` (configuration-extension images, unpacked to `/.extra/confext/`). If we add the two `.raw` suffixes here too, `stage_uki_on_esp` would faithfully stage everything the spec allows in `extra.d`, and we'd avoid coming back to patch this filter again the next time a feature drops sysext/confext images in there.

frhuelsz (approved):

